PR #54, opened 2020-01-14
Update README with how-to install the cap'n proto dependency.

hzeller:


alainmarcel:
There was an INSTALL.md file that was describing that, I should have referenced it in the README

Issue #70, opened 2020-01-15
Object parameters should not be by value but reference


hzeller:
For instance, string parameters are currently passed by value

```
bool set_vpiName(std::string data) 
```

This should be
```
bool set_vpiName(const std::string &data) 
```

(or, if we're using c++17, this should actually be `std::string_view` by value. If we aim for C++11, we could use `absl::string_view`, however that would add another dependency, so `const std::string &` is probably best in that case).

alainmarcel:
Fixed by #87

PR #87, opened 2020-01-21
return const pointers for getter methods

### LINKED ISSUE #71: Pointers to other things: can they be const ?
For instance,we have code like this:
```
    expr* get_right_expr() const { return right_expr_; }
```
This will return a pointer that allows to modify state, yet we have the method marked `const`, which indicates that this todes not modify the object ... we can do so however through the backdoor.

So we should make this return a `const expr*`
```
    const expr* get_right_expr() const { return right_expr_; }
```

_iff_ we need writable and non-writable versions, we should do so having the same signature twice with and without const; the compiler will pick the right one in use.
```
    const expr* get_right_expr() const { return right_expr_; }
    expr* get_right_expr() { return right_expr_; }
```

alainmarcel:
Signed-off-by: Alain <alainmarcel@yahoo.com>
Fixes #71

### DIFF
--- a/model_gen.tcl
+++ b/model_gen.tcl
@@ -241,8 +241,10 @@ proc printMethods { classname type vpi card {real_type ""} } {
     }
     if {$card == "1"} {
 	set pointer ""
+	set const ""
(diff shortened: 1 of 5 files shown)


alainmarcel:
Fixes #70

PR #115, opened 2020-02-07
Remove capnproto limits

alainmarcel:


.travis.yml
```diff
@@ -28,6 +28,8 @@ install:
 
 # Build steps
 script:
+  - sed -i 's/nestingLimit = 64/nestingLimit = 1024/g' third_party/capnproto/c++/src/capnp/message.h
```
hzeller: I think modifying the header in the third_party/ is a bad idea, as we'd typically get the header from the system.

Instead, whenever we use the `RenderOptions`, we should conffigure it accordingly.

alainmarcel: UHDM and Surelog both have their own pull from capnproto (git recursive update).
I do the sed after the recursive update. 
For a manual install I updated the instructions to do the same. The capnp proto on most Ubuntu install is obsolete, that is why I have it as a git pull.

hzeller:
But why not just set the option when you use the struct ?
Even if you subproject change this project, if you modify the submodule you'll get all kinds of issues when updating it, so this is why this is dangerous.

Using the struct as intended, changing the options as needed, is a much more robust approach.

.travis.yml
alainmarcel: I see now how that can be configured in the reader.
I'll try your suggestion.

alainmarcel:
I see now how that can be configured in the reader, I didn't see there were extra optional arguments to the reader constructor until now.
I'll try your suggestion.

PR #126, opened 2020-02-11
Use EXCLUDE_FROM_ALL in the capnp subproject to only build

hzeller:
what we need.

This should fix
  https://github.com/alainmarcel/Surelog/issues/170
and might help
  https://github.com/SymbiFlow/conda-packages/issues/68

Signed-off-by: Henner Zeller <h.zeller@acm.org>

PR #127, opened 2020-02-12
Disable capnproto tests to build.

hzeller:
We don't run them and they generate link issues in some
contexts.


hzeller:
This is a lighter-weight alternative to #126 and good until we've found out how to make that work with older cmakes.

hzeller:
This is the fix suggested by the variable find by @kgugala in https://github.com/alainmarcel/Surelog/issues/170

PR #234, opened 2020-05-14
Directly output to stream

hzeller:
Also, output the normal operation headers to cout, as the Surelog regression tcl script would otherwise complain about things coming to stderr.


hzeller:
We will need this once https://github.com/alainmarcel/Surelog/issues/376 is fixed.
Because when it finds the binary, the tcl `catch` command will still fail because it sees something on stderr (not sure why it is implemented that way, but we have to deal with it :) )

PR #269, opened 2020-06-25
Simplify the way how to add tests. Add option to disable tests.

hzeller:
Added UHDM_BUILD_TESTS so that in other projects that use this
(e.g. Surelog) we can choose to not expliictly run the sub-project's
tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ option(
   WITH_LIBCXX
   "Building with clang++ and libc++(in Linux). To enable with: -DWITH_LIBCXX=On"
   On)
+option(UHDM_BIULD_TESTS "Enable testing." ON)
 
 project(UHDM)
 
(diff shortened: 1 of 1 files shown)


hzeller:
Looks ready to be merged.
Kokoro seemed to have failed but due to unrelated g++ running out of memory.

alaindargelas:
Can you limit the number of parallel jobs in Kokoro? Tim reported the same problem, so that's a recurring issue.
Each UHDM .cpp file takes around 1Gb of RAM, some take up to 1.5Gb in gcc, if I understand kokoro is running on a 16Gb machine, there seems to be quite a few files compiled concurrently in UHDM which would account for the virtual memory issue.

CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ option(
   WITH_LIBCXX
   "Building with clang++ and libc++(in Linux). To enable with: -DWITH_LIBCXX=On"
   On)
+option(UHDM_BIULD_TESTS "Enable testing." ON)
```
alaindargelas, line 20: Typo: BIULD -> BUILD

hzeller: Good catch. Fixed.

hzeller:
Kokoro build addressed in #270 270

PR #274, opened 2020-06-27
vpi_user: handle some properties from the base-class once.

hzeller:
The base class offers a few properties, such as VpiLineNo() or
VpiFile() that can be handled once in vpi_get() and vpi_get_str(),
so that the remaining function body only has to worry about the
'non-standard' properties.

 * Changed vpi_get64()/vpi_get() to handle VpiLineNo() and VpiType().
 * All other properties are output in a similar way as before,
   but in a slightly easier human-browsable switch/case form.
 * Since get64 and get are the same right now, just one one function
   call the other so that we don't duplicate code.
 * vpi_get_str handles VpiFile, VpiName and VpiDefName from the
   base-class, then deals with remaining 'manually'.

There are still a lot of properties that have to be handled
manually, but future work could consider having property-specific
interfaces ('VpiFullNamePropertyContainer') that could further
reduce that.

Also, other parts of the code generation have not been adressed yet, but something like
this might help in the future minimize excessive compile times and improve runtime.

Overall, this reduces the number of branches the compiler has to
encode and the runtime has to consider a lot. Dumping BlackParrot2
got 20% faster.

I verified that the `uhdm-dump` output for Blackparrot2 is identical of the before/after code generator change.

##### Before

```c++
PLI_INT32 vpi_get (PLI_INT32   property,
                   vpiHandle   object) {
  // [... ]

  if (handle->type == uhdmattribute) {
    if (property == vpiLineNo) {
      return ((attribute*)(obj))->VpiLineNo();
    }
  }
  if (handle->type == uhdmattribute) {
    if (property == vpiDefAttribute) {
      return ((attribute*)(obj))->VpiDefAttribute();
    }
  }
  if (handle->type == uhdmattribute) {
    if (property == vpiDefLineNo) {
      return ((attribute*)(obj))->VpiDefLineNo();
    }
  }
  if (handle->type == uhdmattribute) {
    if (property == vpiType) {
      return ((attribute*)(obj))->VpiType();
    }
  }
//...
```

##### After

Note that `vpiLineNo` and `vpiType` don't have to be handled in the other types. Some times only would provide
these properties, they don't even show up anymore in the `if (...)` branches.
```c++
PLI_INT64 vpi_get64 (PLI_INT32 property,
                     vpiHandle   object) {
  // [...]

  // Baseclass-handled properties; all the others still need to be handled
  // separately, but this is a good start.
  switch (property) {
    case vpiLineNo: return obj->VpiLineNo();
    case vpiType: return obj->VpiType();
  }

  // ... all other properties currently handled 'manually' for now
  if (false) {}
  else if (handle->type == uhdmattribute) {
    switch (property) {
      case vpiDefAttribute: return ((const attribute*)(obj))->VpiDefAttribute();
      case vpiDefLineNo: return ((const attribute*)(obj))->VpiDefLineNo();
    }
  }
  else if (handle->type == uhdmconcurrent_assertions) {
    switch (property) {
      case vpiIsClockInferred: return ((const concurrent_assertions*)(obj))->VpiIsClockInferred();
    }
  }
 // [...]
  else if (handle->type == uhdmassertion) {
   switch (property) {
     case vpiStartLine: return ((const assertion*)(obj))->VpiStartLine();
     case vpiColumn: return ((const assertion*)(obj))->VpiColumn();
     case vpiEndLine: return ((const assertion*)(obj))->VpiEndLine();
     case vpiEndColumn: return ((const assertion*)(obj))->VpiEndColumn();
   }
  }
}
```

### DIFF
--- a/model_gen/model_gen.tcl
+++ b/model_gen/model_gen.tcl
@@ -129,8 +129,8 @@ proc printMethods { classname type vpi card {real_type ""} } {
     const BaseClass* parent = this;
     while (parent) {
       if (parent->UhdmType() == uhdmdesign) break;
-      const std::string\\& name = (parent->VpiName() != \"\") ? parent->VpiName() : parent->VpiDefName();
-      if (name != \"\")
+      const std::string\\& name = (!parent->VpiName().empty()) ? parent->VpiName() : parent->VpiDefName();
+      if (!name.empty())
         names.push_back(name);
       parent = parent->VpiParent();
     }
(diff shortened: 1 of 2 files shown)


alaindargelas:
I had to undo this pull request.
All vpiSize are lost in the dump.
When changing UHDM, make a local copy of the changes in Surelog, recompile Surelog and run Surelog regressions, if passing, then submit the UHDM change.
Surelog regression is much more comprehensive than UHDM unit tests. And we don't want to replicate Surelog regression here.
I'm surprised the BlackParrot run didn't generate diffs.

hzeller:
Interesting. Will investigate and come back to you, thanks.

PR #277, opened 2020-06-27
vpi_user: handle some properties from the base-class once.

hzeller:
Redo of pull #274

Tested with Surelog: all regression tests pass.

The base class offers a few properties, such as VpiLineNo() or
VpiFile() that can be handled once in vpi_get() and vpi_get_str(),
so that the remaining function body only has to worry about the
'non-standard' properties.

 * Changed vpi_get64()/vpi_get() to handle VpiLineNo() and VpiType().
 * All other properties are output in a similar way as before,
   but in a slightly easier human-browsable switch/case form.
 * Since get64 and get are the same right now, just one one function
   call the other so that we don't duplicate code.
 * vpi_get_str handles VpiFile, VpiName and VpiDefName from the
   base-class, then deals with remaining 'manually'.

There are still a lot of properties that have to be handled
manually, but future work could consider having property-specific
interfaces ('VpiFullNamePropertyContainer') that could further
reduce that.

Overall, this reduces the number of branches the compiler has to
encode and the runtime has to consider a lot. Dumping BlackParrot2
got 20% faster.

Difference to the #274 is, that we don't if/else if through the
classes but if () each individually: apparently the same class name
is repeated multiple times in the list iterated through in the list
(separate problem independent of this change; to address separately).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,7 @@ if (UHDM_BUILD_TESTS)
     tests/test-dump.cpp
     tests/listener_elab.cpp
     tests/full_elab.cpp
+    tests/vpi_get_test.cpp
   )
 endif()
 
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,8 @@ debug:
 	$(MAKE) -C build
 
 test: build
-	$(MAKE) -C build UnitTests test
+	$(MAKE) -C build UnitTests
+	cd build && ctest --output-on-failure
 
 test-junit: release
 	cd build && ctest --no-compress-output -T Test -C RelWithDebInfo --output-on-failure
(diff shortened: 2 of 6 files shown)


hzeller:
Difference to the previous approach is that it now doesn't do if / else if / else if
```
  if (handle->type == uhdmattribute) {
   // ...
  }
  else if (handle->type == uhdmconcurrent_assertions) {
   // ...
  }
```

but instead not exlusive if / if / if.

```
  if (handle->type == uhdmattribute) {
   // ...
  }
  if (handle->type == uhdmconcurrent_assertions) {
   // ...
  }
```

Reason is, that might be called with the same `$classname` multiple times which then separates into several switch/case. In the if/else case, they are shadowed. So not doing exclusive if/else fixes the issue seen in the original pull request.

hzeller:
Like always when there is a bug, we need a test.
I've now also added a test, that tests for this situation. With the other implementation the `vpiSize` getting fails, with the change, it works.

PR #281, opened 2020-06-28
Make symbolic scalar parsing work; output scalar symbolic names.

hzeller:
While at it, simplify and speed up VpiValue2String() and String2VpiValue().


hzeller:
(Also, Surelog still works with this change, the print visitor is printing things differently, so that is not affected)

PR #282, opened 2020-06-28
Document NewHandle() memory-leak situation and path forward.

hzeller:
In code generation, replace NewHandle() with NewVpiHandle()
where possible, to see the remaining parts to address.

No functional change otherwise.


hzeller:
ah ok. Removed the comment (but leaving in the other readability improvements).

alaindargelas:
I'm adding the vpi_release_handle call back for the Surelog version.
Someone with access to commercial simulators in interactive session can debug why the vpi_visitor crashes (To debug, remove the VPI_STANDARD ifdef in release_handle function).
Maybe only handles to iterators are allocated and need to be released, and handles to not, but gdb would tell.

PR #284, opened 2020-06-29
Switch on optimization for SymbolFactory.

hzeller:
While we don't use optimization for the bulk of the generated files,
because they are slow to compile, SymbolFactory is quick to compile
with optimization and yields > 4x speedup for SymbolFactory
operations (on my laptop, Blackparrot2 compile spent
2.58s vs 11.8s in SymbolFactory::Make)).

So selectively switch on optimization for that file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,9 +63,19 @@ else()
   set(CMAKE_CXX_FLAGS_DEBUG
       "${CMAKE_CXX_FLAGS_DEBUG} ${TCMALLOC_COMPILE_OPTIONS} -Wall -O0 -g ${MY_CXX_WARNING_FLAGS}"
   )
+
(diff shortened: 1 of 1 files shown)


alaindargelas:
Do we still have a proper full debug build vs optimized build?

hzeller:
Nothing changed in the original set-up: the debug build and release build are **both** non-optimized.

Ideally, I'd like to switch on the optimization for all of release, but apparently that was switched off at so
me point (too long compile times or too much memory ?). This is the `-O0` in [this line in CMakeLists](https://github.com/alainmarcel/UHDM/blob/master/CMakeLists.txt#L67).

So here, I now just selectively switched on optimization for one file that really can use it.

Of course, longer term we should consider switching on optimization in the release build, and I hope to get there by reducing the amount of generated code using base-class tricks such as in #277

PR #285, opened 2020-06-29
Optimization: on for all except a few.

hzeller:
Optimization had been switched off for all code except for
one that really needed optimization.

Let's do it the opposite: compile all files with optimization
but excempt a few that take excessively long to compile. This
also makes it very explicit which can benefit the most by
different code generation.

Compilation time roughly doubles on my machine with this change,
but the runtime also is significant: a uhdm-dump is about 3x
faster (On my machine a dump of blackparrot takes 19 seconds vs. 65 seconds).

Given that with ccache we won't really see increased compilation times
after the first compile, this looks like well worth it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,16 +64,19 @@ else()
       "${CMAKE_CXX_FLAGS_DEBUG} ${TCMALLOC_COMPILE_OPTIONS} -Wall -O0 -g ${MY_CXX_WARNING_FLAGS}"
   )
 
-  # A lot of the generated files are so large that the compiler takes a long
-  # time to compile. Leave that at no optimization for now.
   set(CMAKE_CXX_FLAGS_RELEASE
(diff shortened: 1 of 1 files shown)


hzeller:
(running a Surelog regression with these optimizations switched on runs > 15% faster (38min vs. 46min user time on my laptop))

alaindargelas:
One file compiles in 3 minutes, that is not really tolerable especially when tweaking the models often (Which is about to start again for the Simulation models). The 2nd longest takes 2 mins.
Can you also suppress the optimization for the long running? I don't know which one it is.

hzeller:
For tweaking, can you just compile in Debug mode ? That is much faster.

hzeller:
The long runnings are right now probably the ones in the set, but they are compiled with `-O1`. Maybe switching them to `-O0` will help there.

hzeller:
In general, my intent is to find out what makes these files slow to compile and adapt the code generation that this is not a problem anymore...

hzeller:
mmh, wait, in debug mode now the optimization is also overriden with `-O1` for these files. Let me change that.

hzeller:
#286

PR #287, opened 2020-06-29
Make sure the Tcl script does not stop if capnp emits a warning.

hzeller:
The default Tcl exec has the unfortunate feature to fail when
the command in question outputs on stderr, switch this off.


hzeller:
In my case, I was symbolic linking UHDM to Surelog and capnp was confused about the path and emitted a warning. Not something that should stop us from continuing the build.

PR #308, opened 2020-07-17
Add missing generated src from CMakeLists

hzeller:
Also, sort all the generated sources, so that it is easier to pinpoint missing ones.


hzeller:
The first change sorts the existing files, the second one adds the new ones.

PR #362, opened 2020-10-25
Support for CI Builds on GHA

hs-apotell:
Support for CI Builds on GHA

* Includes Linux, MSYS+GCC and Windows+MSVS
* Fixed a few compile errors to support all configurations
* Added a few additional files/folders to ignore
* Don't assume LF for all editors across platforms
* Updated Makefile to support parallel builds

PR #363, opened 2020-10-26
ci: rework 'main' workflow

umarcor:
Coming from #362

In the linux-gcc job, global envvars are used instead of `$GITHUB_ENV`. That's just an style change.

In the msys2-gcc job, several changes are applied:

- MINGW64 is used instead of MSYS. See https://www.msys2.org/wiki/Creating-Packages/#general-information.
  - Therefore, `mingw-w64-x86_64-*` versions of cmake, ninja and gcc are installed.
- Instead of inheriting the environment (which is likely to produce problems), envvars are defined in `~/.bashrc`.
- Step "Show shell configuration" is removed, because it was failing for no obvious reason.
- Build and Test are split in two steps.

Currently the msys2-gcc job is failing because cmake seems to try using `cmd.exe`: https://github.com/umarcor/UHDM/runs/1306520522?check_suite_focus=true#step:6:32

### DIFF
--- a/.github/mingw-w64-UHDM/PKGBUILD
+++ b/.github/mingw-w64-UHDM/PKGBUILD
@@ -1,0 +1,38 @@
+_realname=UHDM
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=ci
+pkgrel=1
+pkgdesc="DESCRIPTION (mingw-w64)"
+arch=('any')
+url="URL"
+license=('license')
+depends=("${MINGW_PACKAGE_PREFIX}-gcc-libs")
+makedepends=("${MINGW_PACKAGE_PREFIX}-gcc"
+             "${MINGW_PACKAGE_PREFIX}-cmake"
+             "${MINGW_PACKAGE_PREFIX}-ninja"
+             "${MINGW_PACKAGE_PREFIX}-zlib"
+             "${MINGW_PACKAGE_PREFIX}-tcl")
+
+build() {
+  export CC=gcc
+  export CXX=g++
+  export PREFIX=/usr/include
+  export CMAKE_GENERATOR=Ninja
+  export NO_TCMALLOC=On
+
+  cd ${srcdir}/../../../
+
+  make release
+}
+
+package() {
+  cd ${srcdir}/../../../
+  make DESTDIR="${pkgdir}" install
+  make DESTDIR="${pkgdir}" test_install
+}
+
+test() {
+  cd ${srcdir}/../../../
+  make test
+}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,13 +5,21 @@ on:
   pull_request:
 
 jobs:
+
+
   linux-gcc:
     runs-on: ubuntu-latest
+
     defaults:
       run:
         shell: bash
+    env:
+      CC: gcc-9
+      CXX: g++-9
+      PREFIX: /usr/include
 
     steps:
+
     - name: Install dependencies
       run: |
         sudo apt-get update -qq
@@ -22,12 +30,6 @@ jobs:
       with:
         submodules: recursive
         fetch-depth: 0
-
-    - name: Configure shell
-      run: |
-        echo 'CC=gcc-9' >> $GITHUB_ENV
-        echo 'CXX=g++-9' >> $GITHUB_ENV
-        echo 'PREFIX=/usr/include' >> $GITHUB_ENV
 
     - name: Show shell configuration
       run: |
@@ -45,57 +47,6 @@ jobs:
         sudo make install
         make test_install
 
-  msys2-gcc:
-    if: false # disable for the time being until we figure out the reason for slow builds and intermittent link issues
-    runs-on: windows-latest
-    defaults:
-      run:
-        shell: msys2 {0}
-
-    steps:
-    - name: Setup Msys2
-      uses: msys2/setup-msys2@v2
-      with:
-        msystem: MSYS
-        update: true
-        install: make cmake ninja gcc
-      env:
-        MSYS2_PATH_TYPE: inherit
-
-    - name: Configure Git
-      run: git config --global core.autocrlf input
-      shell: bash
-
-    - name: Git pull
-      uses: actions/checkout@v2
-      with:
-        submodules: recursive
-        fetch-depth: 0
-
-    - name: Configure shell
-      run: |
-        echo 'CC=gcc' >> $GITHUB_ENV
-        echo 'CXX=g++' >> $GITHUB_ENV
-        echo 'PREFIX=$PWD/install' >> $GITHUB_ENV
-        echo 'CMAKE_GENERATOR=Ninja' >> $GITHUB_ENV
-        echo 'NO_TCMALLOC=On' >> $GITHUB_ENV
-
-    - name: Show shell configuration
-      run: |
-        env
-        where cmake && cmake --version
-        where make && make --version
-        where ninja && ninja --version
-        where tclsh && echo 'puts [info patchlevel];exit 0' | tclsh
-        where $CC && $CC --version
-        where $CXX && $CXX --version
-
-    - name: Build & Test
-      run: |
-        make release
-        make test
-        make install
-        make test_install
 
   windows-msvc:
     runs-on: windows-latest
@@ -105,6 +56,7 @@ jobs:
         shell: cmd
 
     steps:
+
     - name: Install Core Dependencies
       run: |
         choco install -y make
--- a/.github/workflows/msys2.yml
+++ b/.github/workflows/msys2.yml
@@ -1,0 +1,34 @@
+name: 'msys2'
+
+on:
+  push:
+  pull_request:
+
+jobs:
+
+  windows-msys2:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+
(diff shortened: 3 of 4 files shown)


hs-apotell:
The referenced build is showing compile issue. This has not been my experience. This might help - https://github.com/hs-apotell/UHDM/actions/runs/327034509. This is the last MSYS build on the same branch that got merged a few hours back. I had the MSYS build going until it hit a link issue at the end. I will spend some time investigating the link issue.

Looks like you are missing some file edits from last merge. Make sure to rebase. The gha-ci PR included code/script changes as well and not just the workflow file.

Also, doesn't the `make install` belong in Build action and not in Test? You have it currently in the latter in this change.

hs-apotell:
Here's the most recent run where MSYS got past the step that's failing in this repo - https://github.com/hs-apotell/UHDM/runs/1306878493.

P.S. The build might still fail but its not blocked on the same errors as in yours.

hs-apotell:
Link issue resolved. Successful UHDM MSYS build - https://github.com/hs-apotell/UHDM/runs/1306878493?check_suite_focus=true

umarcor:
> Looks like you are missing some file edits from last merge. Make sure to rebase. The gha-ci PR included code/script changes as well and not just the workflow file.

I did rebase, but there were some issues with my previous runs. I now created a PKGBUILD file, as commented in Surelog's issue. As you can see in https://github.com/umarcor/UHDM/runs/1306999718?check_suite_focus=true#step:5:191 it is failing at the UHDM model generation step. Compare with https://github.com/hs-apotell/UHDM/runs/1306878493#step:7:130. Any guess about how to fix that? Is it part of capnp or is it failing right after that?

hs-apotell:
If all the code is the same, then the difference has to be in the workflow file. The only major difference I see in your version (that doesn't work, [here](https://github.com/alainmarcel/UHDM/pull/363/checks)) and mine (which works, [here](https://github.com/hs-apotell/UHDM/actions/runs/328294710)) is the MSYSTEM.  Could you try changing the MSYSTEM to MSYS instead of MINGW64?

At least to get a confirmation on whether that's an issue or not. If not anything that will give us some direction to investigate.
`Permission denied` isn't saying much about _what was attempted_.

umarcor:
> Could you try changing the MSYSTEM to MSYS instead of MINGW64?

Unfortunately, I can't, because that's the point. Binaries need to be built in either MINGW64 or MINGW32. See https://www.msys2.org/wiki/Creating-Packages/#general-information:

> **msys software (from the msys repository) is software that depends on msys-2.0.dll** and is very similar to Cygwin software (which is a POSIX emulation layer for Windows). **Native Windows software (from this project's perspective) is software that doesn't depend on msys-2.0.dll**, and links dynamically to the highly compatible msvcrt.dll.

So, there are multiple differences between MSYS and MINGW64:

- The version of GCC. 
- ` -- Building for: Ninja` is not shown in the MSYS run.
- OpenSSL and ZLIB are not found in the MSYS run.

hs-apotell:
I found the root cause. As you identified earlier, cmake is using `cmd` to invoke the script -
```
cmd.exe /C "cd /D D:\a\UHDM\UHDM && tclsh D:/a/UHDM/UHDM/model_gen/model_gen.tcl D:/a/UHDM/UHDM/model/models.lst D:/a/UHDM/UHDM/build"
```
Here's the snippet that calls capnp executable from the tcl script -
```
if { ($tcl_platform(platform) == "windows") && (![info exists ::env(MSYSTEM)]) } {
  exec -ignorestderr cmd /c "set PATH=$capnp_path;%PATH%; && cd /d [project_path]/src && $capnp_path/capnp.exe compile -oc++ UHDM.capnp"
} else {
  exec -ignorestderr sh -c "export PATH=$capnp_path:\$PATH; $capnp_path/capnp compile -oc++:. [project_path]/src/UHDM.capnp"
}
```
Based on this it will spawn a `sh` shell and in that process adds the build output directory to the PATH environment variable. As per the log in the build, 
```
capnp_path = D:/a/UHDM/UHDM/build/third_party/capnproto/c++/src/capnp
```
which isn't a legal Unix path and so the process doesn't find the necessary files. The path ends up looking this -
```
PATH=D:/a/UHDM/UHDM/build/third_party/capnproto/c++/src/capnp:$PATH
```
I made some changes to the script to reproduce the issue locally and report the underlying cause. It reports the following errors -
```
c++: no such plugin (executable should be 'capnpc-c++')
```
Now the big question is why is there are mixup of `cmd` and `sh`. As per cmake docs [here](http://manpages.ubuntu.com/manpages/bionic/man7/cmake-generators.7.html) building for MinGW should use `cmd` shell with `mingw32-make` as alternative for make. If we wish to use `sh` shell, the generator has to `MSYS Makefiles` but then the environment has to be MSYS as well. I am not sure what role 'MSYSTEM' plays in all this. There is no specific detail on what it uses for Ninja, so my guess would be that it is based on the host OS and not MSYS or MinGW.

Do we want to switch to using `MinGW Makefiles` for this configuration? But then we need to switch to using `cmd` shell, as per cmake docs.

People discussing similar issues -
https://github.com/msys2/MINGW-packages/issues/2701#issuecomment-316066071

umarcor:
> Now the big question is why is there are mixup of `cmd` and `sh`. As per cmake docs [here](http://manpages.ubuntu.com/manpages/bionic/man7/cmake-generators.7.html) building for MinGW should use `cmd` shell with `mingw32-make` as alternative for make. If we wish to use `sh` shell, the generator has to `MSYS Makefiles` but then the environment has to be MSYS as well. I am not sure what role 'MSYSTEM' plays in all this. There is no specific detail on what it uses for Ninja, so my guess would be that it is based on the host OS and not MSYS or MinGW.

Paths can be converted to the Unix style with `cygpath -u -a 'C:\'`. Conversely, `cygpath -m -a '/c'` will produce a Windows style path.

Anyway, I will ask about cmake's specific behaviour on MSYS2.

umarcor:
[October 26, 2020 8:27 AM](https://gitter.im/msys2/msys2?at=5f967a496c8d484be2ab3f08)

> According to http://manpages.ubuntu.com/manpages/bionic/man7/cmake-generators.7.html, there are two options for using cmake with MSYS2:
>
> - "a MSYS shell prompt and using make".
> - "a Windows command prompt with MinGW in the PATH and using mingw32-make".
>
> Isn't it possible to use mingw-w64-cmake in a MINGW64 shell and without cmd.exe? Which make should be used, make or mingw32-make?

> under MSYS always use make with "MSYS Makefiles" generator

> I want to use cmake in a PKGBUILD file for MINGW32/MINGW64. Is that considered "MSYS"?

> MINGW32/64 shell is some kind of MSYS shell

> or use Ninja generator with ninja

> I've updated the templates recently to use Ninja: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball

hs-apotell:
The question wasn't really answered in that conversation. With Ninja in play, cmake uses `cmd` shell which means all the paths will be windows style. These paths cannot be used build legal environment variables, especially not in context of `PATH`.

One potential solution is to treat `MSYSTEM == MINGW64` as more like Windows environment and less like Unix environment. At the moment, we check only for the presence of `MSYSTEM` variable in the environment. May be we need to add another condition to check if its set to `MINGW32` or `MINGW64`. Evaluating this change locally now.

umarcor:
I feel that this might be an important hint: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball#L42

hs-apotell:
I am not sure I follow where you are going with this.

umarcor:
> I am not sure I follow where you are going with this.

MSYS2/MINGW does conversions to use Unix style paths in the shell, but pass windows style paths to the binaries. `MSYS2_ARG_CONV_EXCL` is an environment variable that prevents some paths from being converted. OTOH, https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball is a template for building cmake based projects using Ninja, which is our use case. I'd say that zhttps://github.com/msys2/MINGW-packages/blob/master/mingw-w64-PKGBUILD-templates/PKGBUILD.CMake-tarball#L42 is the recomended command for "configuring" cmake.

hs-apotell:
> `MSYS2_ARG_CONV_EXCL` is an environment variable that prevents some paths from being converted. 

But we want the opposite option. We want a Windows style path to be converted to Unix style path. Moreover, I am not sure how this will even work for a shell command that's being invoked from within a script. This will prevent (or otherwise) help convert paths back and forth for what's passed down to cmake or what cmake itself passes down. But for exec commands from within scripts, it's likely a stretch.

umarcor:
For now, I applied the `cygwin -u -a` patch to the TCL script, and it seems to go ahead. It broke the linux job, because the fix needs to be applied on MSYS2 only. I'll check that later.

umarcor:
Work on MINGW packages moved to https://github.com/hdl/MINGW-packages/tree/UHDM/mingw-w64-UHDM.

umarcor:
Ref msys2/setup-msys2#84

hs-apotell:
We could try switching to MSYS Makefiles and evaluate.

However, this would be like sliding the problem under the rug. Given that the same build works as expected locally in the same environment, why would a build on CI be any different. There has to be a more logical/rationale explanation of the outcome on the CI node. If Ninja is the "trouble maker" we still need to understand as to why is it behaving any different between the two machines.

umarcor:
Ref: https://github.com/msys2/MINGW-packages/issues/7240

PR #387, opened 2020-12-22
Partially revert #381

### LINKED ISSUE #386: Windows build broken
@alaindargelas the issue with capnproto you mentioned during the telco seems to be caused by #381 - it seems to be updating the capnproto submodule.

tgorochowik:
This restores the previous capnproto version (to fix the CI build).

Closes #386


tgorochowik:
@alaindargelas the CI passes for everything so assuming the module bump was accidental (or isn't really needed) we should merge this

PR #416, opened 2021-04-28
Fix Tcl indentation and remove trailing whitespace.

hzeller:


mithro:
It looks like there is a tcl re-formatter in prjxray - https://github.com/SymbiFlow/prjxray/blob/master/utils/tcl-reformat.sh

PR #469, opened 2021-08-09
Don't clobber source directory with test outputs.

hzeller:
Tests should only access directories explicitly made available
for temporary files by the test harness. Add a simple function
that is checking for common environment variables TEST_TMPDIR,
TMPDIR and then fallback to /tmp.


hzeller:
converting to draft right now until CI issues are settled.

PR #487, opened 2021-09-09
Revert "Put all generated code in the build/ directory."

alaindargelas:
Reverts chipsalliance/UHDM#485

Please fix the Surelog build issues.

PR #488, opened 2021-09-09
Install all uhdm related headers in flat $PREFIX/include/uhdm/*.h

### LINKED ISSUE #484: Cleanup of header installation
Currently, we have two include directories that are installed:

  * `${INSTALL_PREFIX}/include/uhdm/headers`  _the generated headers_
  * `${INSTALL_PREFIX}/include/uhdm/include`   _vpi_stuff_
  
... and then there is also `uhdm.h` installed twice, in `uhdm/uhdm.h` and `uhdm/headers/uhdm.h` (#197)

The names of 'include/' and 'headers/' is very generic, so it does not convey any meaningful subdivision of functionality - they just happen to be that way because one set of files is generated.
From the users' perspective it is an arbitrary choice.

So I propose we should just have _one_ `${INSTALL_PREFIX}/include/uhdm/*.h` with all the headers.
For users that don't have both `-Iuhdm/include` and `-Iuhdm/headers` in their include path and use things like `#include "expect_stmt.h"`,  but use full paths such as `#include "uhdm/headers/expect_stmt.h"`, this will be a bit of a disruption to change it to `#include "uhdm/expect_stmt.h"`, but I'd argue that since the user-base is still smallish, this will be less of a problem now than later.

WDYT @alaindargelas @MikePopoloski  @hs-apotell ?

hzeller:
**DON'T MERGE INTO SURELOG YET**

This removes the arbitrary distinction between relatively
meaningless names header/ and include/

Adapt all code that used the previous paths in their includes
to point to the new location.

Install location is presumed to be a 'system location', i.e. the distributed includes are used in angled brackets,
such as `#include <uhdm/uhdm.h>`

This likely cause one-time disruption for current users that directly included things from `uhdm/headers/foo.h` as these are now in `uhdm/foo.h`; currently the user-base is small enough to be a limited disruption.

While at it, install also the UHDM.capnp in `$PREFIX/lib/uhdm`
for other tools using different languages to pick up just the
schema.

Fixes #484

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MY_CXX_WARNING_FLAGS}")
 
+# Directory where code is generated into.
+set(GENDIR ${CMAKE_CURRENT_BINARY_DIR}/generated)
+
 if(MSVC)
   set(CMAKE_CXX_FLAGS_DEBUG
       "${CMAKE_CXX_FLAGS_DEBUG} ${TCMALLOC_COMPILE_OPTIONS} /W3 /bigobj ${MY_CXX_WARNING_FLAGS}"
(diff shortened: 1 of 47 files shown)


hzeller:
... we'll only get this in once the previous change #487 is in again.

hzeller:
Once this is merged, I have another change that then updates Surelog to use the new includes (e.g. `#include <uhdm/uhdm.h>`) and updates the UHDM submodule.

But till then, merge this into UHDM (so that I can refer to head in the Surelog change), but do not update submodule in Surelog alone (as that will not compile without the other change)

Issue #483, opened 2021-09-07
Consider building against a fixed Capn Proto target


MikePopoloski:
It would be nice if UHDM targeted a specific Capn Proto release instead of living at head via a submodule, for interoperability purposes. Each commit to the upstream repo can potentially break any consumer that isn't compiled against exactly the same commit, which increases in likelihood if there are other users of UHDM besides Surelog.

It seems easier to just rely on whatever version is installed on the system (using FindPackage in cmake). On Windows users will need to build the lib themselves but that's not really any different than how it works today.

hzeller:
+1, this looks like someone with CMake-fu should be able to implement that easily.

Looking at debian packages, it looks like it comes with pkg-config packages
https://packages.debian.org/sid/amd64/libcapnp-dev/filelist

For windows users: Maybe there is also a similar devel package on chocolatey ?

hzeller:
Also, this involves to generate the capnproto outputs in the CMakefile instead of the tcl-script.

PR #491, opened 2021-09-09
Use native capnproto when building UHDM.

hzeller:
Work in progress. I don't know yet how I can tell CMake that there is a dependency to first create the `UHDM.capnp` file before running `capnp_generate_cpp()`

This is working towards #483 

@hs-apotell can you help ? I think you have the most cmake-fu of us all

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,9 +25,6 @@ project(UHDM)
 cmake_policy(SET CMP0091 NEW)
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
-set(BUILD_TESTING OFF CACHE BOOL "Don't build capnproto tests")
-add_subdirectory(third_party/capnproto EXCLUDE_FROM_ALL)
-
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
@@ -76,9 +73,7 @@ else()
 
 endif()
 
-# model_gen generated
-set(model-GENERATED_SRC ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp.h
-                        ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp.c++)
+find_package(CapnProto)
 
 # All the files the generator depends on.
 file(GLOB yaml_SRC ${PROJECT_SOURCE_DIR}/model/*.yaml)
@@ -86,12 +81,10 @@ file(GLOB yaml_SRC ${PROJECT_SOURCE_DIR}/model/*.yaml)
 file(GLOB templates_SRC ${PROJECT_SOURCE_DIR}/templates/*.h
      ${PROJECT_SOURCE_DIR}/templates/*.cpp)
 
-foreach(header_file ${model-GENERATED_SRC})
-  set_source_files_properties(${header_file} PROPERTIES GENERATED TRUE)
-endforeach(header_file ${model-GENERATED_SRC})
-add_custom_target(GenerateCode DEPENDS ${model-GENERATED_SRC})
+set(model-GENERATED_UHDM ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp)
+set_source_files_properties(${model-GENERATED_UHDM} PROPERTIES GENERATED TRUE)
 add_custom_command(
-  OUTPUT ${model-GENERATED_SRC}
+  OUTPUT ${model-GENERATED_UHDM}
   COMMAND tclsh ${PROJECT_SOURCE_DIR}/model_gen/model_gen.tcl
           ${PROJECT_SOURCE_DIR}/model/models.lst ${CMAKE_CURRENT_BINARY_DIR}
           ${CMAKE_CURRENT_BINARY_DIR}
@@ -103,6 +96,18 @@ add_custom_command(
           ${PROJECT_SOURCE_DIR}/templates/UHDM.capnp
           ${yaml_SRC}
           ${templates_SRC})
+
+set(model-GENERATED_SRC ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp.h
+                        ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp.c++)
+foreach(header_file ${model-GENERATED_SRC})
+  set_source_files_properties(${header_file} PROPERTIES GENERATED TRUE)
+endforeach(header_file ${model-GENERATED_SRC})
+
+add_custom_command(
+  OUTPUT ${model-GENERATED_SRC}
+  COMMAND capnp compile -oc++ ${model-GENERATED_UHDM}
+  DEPENDS ${model-GENERATED_UHDM})
+add_custom_target(GenerateCode DEPENDS ${model-GENERATED_SRC})
 
 set(uhdm-GENERATED_SRC
     ${CMAKE_CURRENT_BINARY_DIR}/src/ExprEval.cpp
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -104,6 +97,21 @@ add_custom_command(
...
+                        ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp.c++)
+foreach(header_file ${model-GENERATED_SRC})
+  set_source_files_properties(${header_file} PROPERTIES GENERATED TRUE)
+endforeach(header_file ${model-GENERATED_SRC})
+
+# Q: how can I tell capnp_generate_cpp that UHDM.capnp is generatede first?
+# DEPENDS does not work.
+capnp_generate_cpp(CAPNP_SOURCE CAPNP_HEADER
+                  ${CMAKE_CURRENT_BINARY_DIR}/src/UHDM.capnp
+                  DEPENDS ${model-GENERATED_UHDM})
```
hzeller, line 111: It is not clear to me how to tell cmake that we really first have to run the command that gernrates the UHDM before this can be run.

hs-apotell: `capnp_generate_cpp()` is a function and there is no explicit mechanism in cmake to add dependency between targets and function calls. The only alternative is to wrap the function call itself in a custom command and use that command to setup dependencies. Here's an [example](https://stackoverflow.com/questions/26024235/how-to-call-a-cmake-function-from-add-custom-target-command)

Also, note that function execution happens at build configuration time and will be delayed until the end automatically if the outputs of it have no explicit dependencies.

Hope this helps!!

hzeller: Thanks. After fiddling with a `-P` sub-call to cmake, which resulted in its own set of problems (e.g. can't use `FindPackage(CapnProto)` in there), I resorted in making it just a `add_custom_command()`

hzeller:
@hs-apotell What would be the necessary install targets to install external capnproto on Windows ?
I did [some guessing](https://github.com/chipsalliance/UHDM/pull/491/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3) in the CI file, but that didn't work out so well :)

hzeller:
**MSYS2**: Tried mingw64 on msys and that has a capnproto, though simple `capnproto` as package didn't work, just with oddly specific [`mingw-w64-x86_64-capnproto`](https://packages.msys2.org/package/mingw-w64-x86_64-capnproto?repo=mingw64) - am I doing something wrong ?

**MSVC** This does seem to go past installing `capnproto` with `choco`, however that does not seem to install the cmake package in a way that CMake can `find_package()` it, and that is the only package [tagged with capnproto](https://community.chocolatey.org/packages?q=tag%3Acapnproto) there, so no extra 'dev' package of sorts. Do you know where to get the relevant cmake files in that environment @hs-apotell  ?

hzeller:
I think this is mostly ready for review, but question about the Windows CI is open - how to install cmake capnproto package files.

hs-apotell:
I would recommend not doing it this way because of the platform differences. Unlike Linux there is no _default_ location for installations on windows. Even if you manage to get the package files, you will still have to tell CMake where to find the installed package.

Alternatively, you could download the release directly from the GitHub repo. https://github.com/marketplace/actions/get-latest-release

Check the above link. You can do this consistently across platforms rather than specializing the solution for each platform.

P.S. I am on a small screen at the moment and very possible I misread your proposed solution. But I wanted to get this out to avoid losing crucial time and effort.

hzeller:
mmh, but isn't the purpose of the cmake file that is found with `find_package()` to tell the cmake where the system-specific installation of the library and headers are ?

hs-apotell:
In theory, Yes. But Windows doesn’t have that “system specific installation” location for CMake to piggyback on. We run into this problem constantly with other libraries too and in some cases like Qt it registers the installation location in the registry which then CMake can look for. Every such library has its custom solution and those custom solution has now become part of CMake installation on windows. You can look up the logic for ‘find_package’ in your local CMake installation folder.

hzeller:
Ok, so how can we fix that for windows ? I only have a hazy idea how that could be done as I am not familiar with Windows.
Would you like to help out and fix the continuous build on Windows after this is submitted ?

hs-apotell:
I can look into it tonight in your PR. Hold off the PR from merging.

hs-apotell:
Hey, Just investigated what you are attempting to do and what you have done. Unfortunately, this solution is not portable across platforms, especially for Windows. CapnProto doesn't publish any release packages i.e. prebuilt binaries for any platform. Their source repository has releases but they are packaging the entire codebase and releasing it as source archive not platform specific binaries. Yes, it gets a whole lot more complicated to support matrix of supported platforms, compilers, and their individual versions. However, some package manager (like apt-get, pacman, and brew) do publish prebuilt binaries but all those solutions work only for unix-like platforms, not Windows. This page [here](https://capnproto.org/install.html) provides the installation details. For Windows, the only solution is to pull the repository and build it locally. Regardless of whether we do that build as a submodule or we do it as an ExternalProject, it has to be done locally.

Is there any specific reason, why you want to use prebuilt libraries and not build the project as part of the bigger project itself? If build time is your concern, we can solve that by forking the source repository and publishing the built binaries our self. We should be doing this regardless for all the dependent projects anyways at some point. Submodules are great but only at development stage. Once we get to production releases, every project should be capable of standing on its own. 

Given portability concerns, I wouldn't recommend moving ahead on this path.

hzeller:
Ok, so for Windows we'd have to wait until the package managers do distribute the associated cmake package files, as this is what would be needed for platform independent assembly of larger subsystems. From what I gather, the `choco` system tries to go in that direction, this is why I hoped that it was as simple as adding that dependency.

As to why. It got triggered by the request in #483 but in general this makes sense and is common practice: have one project import all the dependencies from the system that is already there, and has a separate release cycle.
It is a little bit icky that we have to install a capnp library together with the udhm library while the former should be just coming from the system. 

Having a project that compiles and installs a slightly different version of a standard library is usually the source for headaches and hard-to-debug issues. Problems arise not only from a potential DLL-hell situation (luckily, we're building with sttic libs here), but the problems that happen with diamond dependncies: when the same binary uses multiple libraries, each pulling in capnproto dependencies, but linking slightly different versions of the same library. Not only bloats that the binaries, but also can result in subtle problems if the linker settings are not right and symbols mix.

Probably for now as a first stability measure, we should tie the capnproto submodule to a common stable release (0.0.8 or 0.0.9?). I think right now the submodule mostly tracks head, which is probably not the best.

Anyway, will convert this to a draft for now to make sure that it doesn't get merged accidentally, but leaving open for discussion and future reconsideration.

hs-apotell:
> the choco system tries to go in that direction, this is why I hoped that it was as simple as adding that dependency.

Yes, choco is an option but AFAIK (and I could be wrong here) developers don't use choco on local development machines. It is used primarily for CI workflows. Among the many reasons as to why - any installation customization (i.e. if I prefer to install in one location over other) requires a licensed paid version. CI machines are _one time use_ so couldn't care less where the stuff goes as long as its consistent but development machines need regular maintenance and up keeping. I wouldn't bank on this solution for local development.

> It is a little bit icky that we have to install a capnp library together with the udhm library while the former should be just coming from the system.

I wouldn't open a debate on this but I, sort of, disagree on this. The rules are different for static vs. shared libraries. If a project has an explicit dependency on a static library that needs to be linked in then it should publish that dependency itself. That allows edited/modified versions of the library to be made available without overhead. For shared modules, use the available system version. Unlike Java, C++ has far too many compile/link options that could change the behavior of the generated output. Unless one is compiling everything on the box that is likely to use the binary, assume it non-portable. Prebuilt static library is _just_ going to work sounds great on paper but hardly ever works out well in practice. Publishing binaries also allows for hot-patching releases if and when needed. Prebuilt tool binaries are a different story all together. Moreover, versioned/side-by-side static library installations isn't always a smooth path on any platform.

> In #483, @MikePopoloski specifically states "On Windows users will need to build the lib themselves but that's not really any different than how it works today."

This is prioritizing one platform over other. I would highly discourage anything moving in this direction. Been on enough projects (and ultimate disasters) where focusing on one platform ultimately forces abandoning other platform because of maintenance overhead and manpower limitations. If we are building on one, lets build on other too. If anything we should look into having stable UHDM releases so Surelog can just build against a UHDM _release_ rather than as a submodule build. Same goes for antlr as well.

> we should tie the capnproto submodule to a common stable release (0.0.8 or 0.0.9?)

Yes, agreed. Knowing explicitly what version of a library is used always helps. Final binaries should also publish such version information that is easily accessible.

> Anyway, will convert this to a draft for now to make sure that it doesn't get merged accidentally, but leaving open for discussion and future reconsideration.

Sounds like an _excellent_ plan for now.

Issue #480, opened 2021-09-02
Garbage collect UHDM when saving


alaindargelas:
A lot of temporary UHDM object are created while performing the Elaboration in Surelog.
These objects should not be serialized on disk, only the objects that are accessible from the designs handles should be serialized.

hzeller:
Can you give me entry points (links to code on github) where the save is happening and an examples of where a temporary object is created, or a temporary object is abandoned (I could imagine 'abandoning' might happen if we have some node like `a + b` which then is replaced by the sum).

Do you have an example of a particularly _large_ test case that allocates lots of memory that I can run on my machine to play around with ideas ? I suspect you have some of these in the Surelog test-suite, so if you can give me a complete command line how to run one of these, would be very helpful.

My initial idea would be to intrinsic reference counting (i.e. the counter is inside the UHDM object and it `delete this;` when it goes to zero), which would make that simple. It would require that we don't have circles back, but I'd suspect everything is represented as DAG.

Anyway, need to stare at some examples to form some opinion towards a solution.

alaindargelas:
I created a new unit test that demonstrate this with the minimum set of objects,
a design, a module part of the design and a module not part of the design.
The latter is the one that can be seen as "temporary", it is serialized but it should not. 

```
~/UHDM$ ./build/bin/test_garbage_collect 
Make design
VISITOR:
design: (design1)
|vpiName:design1
|uhdmtopModules:
\_module: M1 (M1), parent:design1
  |vpiTopModule:1
  |vpiDefName:M1
  |vpiFullName:M1

Save design
Restore design
decompile:
design: (design1)
|vpiName:design1
|uhdmtopModules:
\_module: M1 (M1), parent:design1
  |vpiTopModule:1
  |vpiDefName:M1
  |vpiFullName:M1

decompile:
module: M1 (M1), parent:design1
|vpiTopModule:1
|vpiDefName:M1
|vpiFullName:M1

decompile:
module: M2 (u1)
|vpiDefName:M2
|vpiName:u1

VISITOR:
design: (design1)
|vpiName:design1
|uhdmtopModules:
\_module: M1 (M1), parent:design1
  |vpiTopModule:1
  |vpiDefName:M1
  |vpiFullName:M1

ALL OBJECTS:
OBJECT:
decompile:
design: (design1)
|vpiName:design1
|uhdmtopModules:
\_module: M1 (M1), parent:design1
  |vpiTopModule:1
  |vpiDefName:M1
[...]

PR #493, opened 2021-09-11
Disable garbage collect test for now.

hzeller:
There is some use-after-delete or similar issue going
on; since this is only used in the context of #480,
disable for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,7 +230,7 @@ if (UHDM_BUILD_TESTS)
     tests/vpi_value_conversion_test.cpp
     tests/testErrorHandler.cpp
     tests/test_classes.cpp
-    tests/test_garbage_collect.cpp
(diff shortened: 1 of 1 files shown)


alaindargelas:
In templates/Serializer_save.cpp 

void Serializer::Purge() {
  allIds_.erase(allIds_.begin(), allIds_.end());     <<<< Add this line and merge it, thanks!
<FACTORY_PURGE>
}

alaindargelas:
You can reenable the test after

alaindargelas:
@hzeller, I won't merge or sync uhdm in the next few days since there is a lot of WIP going on there.
Can you merge the line above in the main?

hzeller:
Done ( #496 ). I just make id `allIds_.clear();` as then the container does not have to worry about the iterator range which anyway covers everything.
As soon as the CI is green on that one, I'll just merge it.

PR #500, opened 2021-09-15
Issue #817 Use Python as build scripting tool

hs-apotell:
Issue https://github.com/chipsalliance/Surelog/issues/817 Use Python as build scripting tool

Swap out tcl in favor of Python. Notable changes in transition include the
following:

* Unlike previous tcl based implementation, only the classes account towards
  UHDM ids. This assures that introducing new declarations in existing models
  do not jumble everything.
* Order of the following is now identical to the order of declarations in
  model file
  * Variable & function declarations in header files
  * Order of callbacks in vpi_visitor, VpiListener & Serializer
* Introducing two new outputs
  * VpiListenerTracer - Prints formatted ordered callbacks to assist
     in debugging
  * class_hiearchy - A simple formatted text file that list all models
     in OO hierarchy
* In applicable generated files (uhdm_types, uhdm_forward_decl, containers,
  and alike) the output is sorted
* Each individual script is independently executable

### DIFF
--- a/scripts/ElaboratorListener_h.py
+++ b/scripts/ElaboratorListener_h.py
@@ -1,0 +1,203 @@
+import config
+import file_utils
+
+
+def _generate_module_listeners(models):
+    listeners = []
+    classname = 'module'
+    while classname:
+        model = models[classname]
+
+        for key, value in model.allitems():
+            if key in ['class', 'obj_ref', 'class_ref', 'group_ref']:
+                name = value.get('name')
+                type = value.get('type')
+                card = value.get('card')
+
+                cast = 'any' if key == 'group_ref' else type
+                Cast = cast[:1].upper() + cast[1:]
+                method = name[:1].upper() + name[1:]
+
+                if (card == 'any') and not method.endswith('s'):
+                    method += 's'
+
+                if card == '1':
+                    listeners.append(f'if (auto obj = defMod->{method}()) {{')
+                    listeners.append( '  auto* stmt = obj->DeepClone(serializer_, this, defMod);')
+                    listeners.append( '  stmt->VpiParent(inst);')
+                    listeners.append(f'  inst->{method}(stmt);')
+                    listeners.append( '}')
+
+                elif method in ['Task_funcs']:
+                    listeners.append(f'if (auto vec = defMod->{method}()) {{')
+                    listeners.append(f'  auto clone_vec = serializer_->Make{Cast}Vec();')
+                    listeners.append(f'  inst->{method}(clone_vec);')
+                    listeners.append( '  for (auto obj : *vec) {')
+                    listeners.append( '    enterTask_func(obj, defMod, nullptr, nullptr);')
+                    listeners.append( '    auto* tf = obj->DeepClone(serializer_, this, inst);')
+                    listeners.append( '    ComponentMap& funcMap = std::get<2>(instStack_.at(instStack_.size()-2).second);')
+                    listeners.append( '    funcMap.insert(std::make_pair(tf->VpiName(), tf));')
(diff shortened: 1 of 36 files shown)


hzeller:
Before we switch to another scripting language, I'd _strongly_ suggest to switch to C++ as generator language. It is _much_ easier to maintain (at least for @alaindargelas and me).

scripts/ElaboratorListener_h.py
```diff
@@ -0,0 +1,203 @@
...
+        for key, value in model.allitems():
+            if key in ['class', 'obj_ref', 'class_ref', 'group_ref']:
+                name = value.get('name')
+                type = value.get('type')
+                card = value.get('card')
+
+                cast = 'any' if key == 'group_ref' else type
+                Cast = cast[:1].upper() + cast[1:]
+                method = name[:1].upper() + name[1:]
+
```
hzeller, line 20: Nicely readable section of what happens for different cardinalities and method types!

hs-apotell:
C++ seems overkill for the job. Having to compile some other code first to build a tool that will generate code to be able to build this library... sounds odd. It will force us to release prebuilt binaries for all the supported platforms and configurations - potentially a rather large matrix.

hzeller:
The tools would just be build as part of the build process.

scripts/ElaboratorListener_h.py
hs-apotell: I tried to keep it modular rather than cram everything into a single large script. Hopefully, it would be easier for non-python developers to follow.

alaindargelas:
Nice job, but before I try this, please re-add all the Tcl scripts.
I won't be modifying the Python code, I'm afraid learning a new language does not fit in my schedule at this juncture.
So I'll be maintaining the Tcl scripts and propose you to make the changes in the Python code for a while. 
Also make it so the CMake has a nice way to select the Tcl or the Python generation.

hs-apotell:
@hzeller I might be a little biased on the Python vs. C++ debate. As a portable build tool. I have never had issues with using Python. In my opinion, Python is a good choice for build logic since performance isn't exactly _critical_ and reporting on runtime errors and logging is fairly straight forward. C++ on other end is good for building larger systems and when performance is critical. I like to think of Python as a shell scripting tool like bash but something that works across platforms. The long term maintenance cost of a C++ solution is many fold higher relative to a Python based solution. But surely with the advantage of larger developer pool (even that is debatable these days).

C++ is _the_ common language among all the contributors and I don't think anyone is going to complain about a native solution (fingers crossed). With all that said, a _done right_ C++ solution would be welcome change.

I am going to take the role of a spectator for this debate. I am happy to chime in, if there's a need for an UN-opinionated opinion on something. I would leave the final decision on whether to accept this change or not to _other_ primary contributors and stake holders. Internally, we have maintained this change within our codebase for last 10+ months and we will continue to do so going forward until a _preferred_ solution is in place.

@alaindargelas  Sure thing. I can make the necessary changes.

Please make every possible attempt to coordinate *large* changes to the tcl scripts. On our front, we will try our very best to update the python generator with least possible lag.

Heads up - We are in the process of submitting a follow-up PR for https://github.com/chipsalliance/Surelog/issues/1838 and https://github.com/chipsalliance/Surelog/issues/1836. Expect that to happen in next week or so. The generated output might diverge too far with these changes unless you plan on updating the Tcl scripts as well to follow.

hs-apotell:
Added an option `WITH_PYTHON_GENERATOR` to toggle between Tcl and Python. Default value for it is OFF so using Tcl, by default.

hzeller:
If there are two scripting options, they both should be exercised in the CI. I suggest a build matrix.

hs-apotell:
+1 CI needs to be updated for both UHDM and Surelog.

PR #502, opened 2021-09-16
Fix-up some obj_ref/class_ref mix-ups.

hzeller:
class_refs that refer to obj_def and vice versa.


alaindargelas:
Ooops, the code generator should check for that.

Issue #510, opened 2021-09-20
Add matrix to test different script code generations


hzeller:
Currently, there are two scripts for generating the code. Tcl, which is maintained by Alain as it is his language of choice and an experimental Python script that intends to potentially replace that in the future.
We need to make sure the code generation does not diverge while they are both there. So we need

  * [x] A [matrix](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix) in the CI to run the code generation and tests.
  * [ ] Structural comparison: Testing that the both codes are equivalent is probably hard (different spacing and such), but to show that they are structurally equivalent should be simple: compare the corresponding files and check that they have the same nesting of structure elements  `tr -dc "{}[]()" < some-generated-file.cc`, which then results a string like `(){(){({})(){()}}(){(())[]}(){()(())}}` - compare these 'structure fingerprint'-strings. While this is only a quick smoke test, it will easily detect if there are additional branches etc. in one vs. another.
  * [ ] Also a matrix in Surelog; I think in particular running the regression there with two different backends makes sense.

These tests should be in place before further changes happen in the scripts so that we have the confidence that things are equivalent.

@hs-apotell would you like to take on this task ?

hs-apotell:
Our team is actively working on some of the reported issues to reduce code duplication and making the generated code more modular. Our focus is only on the Python version (we don't have a Tcl expert to really pull off this change in other version).

We expect the change to hit PR stage very soon. I don't think either of these tasks would make much sense once this change hits master.

Moreover, one of the most important difference between the two versions is that Python retains the order of declarations in model_def file while Tcl doesn't. That makes the generated code very different - Functions are ordered differently and the code within each function is ordered differently. A simple regex like what you suggest wouldn't work. We spent a lot of time to compare the generated code between two versions to ensure equivalence and all that was manual. Doing that programmatically would be a whole different challenge.

hzeller:
At least point 1 and 3 for the CI test be implemented.

alaindargelas:
Switch to Python generation completed

PR #513, opened 2021-09-21
 Issue #510: Add matrix to test different script code generations

hs-apotell:
Issue #510: Add matrix to test different script code generations

Updated the workflow to build using both tcl and python generators.

Also, fixing broken build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,10 +6,17 @@ on:
 
 jobs:
   linux-gcc:
+    name: ubuntu-with-${{ matrix.mode }}
     runs-on: ubuntu-latest
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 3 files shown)


hzeller:
It looks like the CMake didn't pick up to run the Python generation environment variable yet - we see the same Tcl log message when generating the code:
https://github.com/chipsalliance/UHDM/pull/513/checks?check_run_id=3666550713#step:7:186

hzeller:
Thanks for working on this!

hs-apotell:
> CMake didn't pick up to run the Python generation environment variable

I thought CMake picks up environment variables automatically. I was wrong. Updated the CMakeLists.txt to compensate for it.

hzeller (approved):
LGTM

PR #537, opened 2021-10-05
Modularize generated code

hs-apotell:
Modularize generated code
- Introducing a new restriction - model's filename should match the
  generated model's class name.
- Serializer implementation is not split into three files -
  Serializer.cpp, Serializer_save.cpp and Serializer_restore.cpp
- capnp.py indexes each model's field in top-down hierarchy rather than
  bottom-up order. This ensures that a field from base class always
  gets the same id regardless of the depth of inheritance.
- classes.py now generates an implementation file for each model. This
  is to reduce the bloat in Serializer related implementation files.
- classes.py also incorporates the logic for cloning. Introduced a new
  function BaseClass::DeepCopy(...) which walks the inheritance
  hierarchy upwards to avoid duplicating the code in each subclass.
  clone_tree_cpp.py is a stub as a result.
- Serialization save/restore now takes advantage of the class hierarchy.

Known Issues
- There is still code in clone_tree.cpp that belongs in model specific
  implementation file.
- Does capnp support 'extends' for structs? Documentation doesn't say
  anything about it. If it does, serializer save/restore logic could
  be lot simpler.
- Ideally, model specific serialization logic should be in respective
  implementation file but the declaration of such a function in the
  header requires including UHDM.capnp.h - which is non-ideal. The
  Reader/Builder classes for each capnp model are nested classes and
  so they can't be forward declared.


hs-apotell:
@hzeller @alaindargelas This change is part one. A follow up PR would tackle the code duplication in vpi_user, vpi_listener and vpi_visitor.

PR #550, opened 2021-10-31
Use precompiled headers to improve build times.

hs-apotell:
Use precompiled headers to improve build times.

Initial numbers look very promising with Windows showing
over 70% improvement on build times.

Gotcha - Certain files have different optimization flags at
the moment and those differ from the ones the precompiled header
is built with. This shows up as an issue with clang on MacOS but
none of the other platforms. Resolution is to disable use of
precompiled headers for those certain files (unfortunately, those
are the ones that can benefit the most with use of precompiled
header).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,14 @@ else()
       ${GENDIR}/src/Serializer_save.cpp
       ${GENDIR}/src/Serializer_restore.cpp
       PROPERTY COMPILE_FLAGS -O0
-   )
+    )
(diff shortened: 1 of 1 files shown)


alaindargelas:
Baseline:
TOTAL ELAPSED | 0s       | 933s     |

After this 2 PR:
|TOTAL ELAPSED | 0s       | 1017s    |

You made the runtime 10% slower with one or both of these changes (That is not going the right direction for the users),
Please systematically make runtime analysis of your changes and report them in the PR.

PR #588, opened 2021-12-18
Transitive group member

alaindargelas:
@hs-apotell, please fix this in your Python version too.
Group checks were not recursively checking group members (Maybe you fixed that already).

### DIFF
--- a/model_gen/model_gen.tcl
+++ b/model_gen/model_gen.tcl
@@ -702,6 +702,48 @@ proc defineType { def name vpiType } {
     return [list $id $define]
 }
 
+proc recurse_generate_group_checker { model } {
+    global $model BASECLASS ALL_CHILDREN GROUP_MEMBERS NAME_TO_MODEL
+    set data [subst $$model]
+    set groupname [dict get $data name]
+    set modeltype [dict get $data type]
+    set checktype ""
+    dict for {key val} $data {
+        if {($key == "obj_ref") || ($key == "class_ref")  || ($key == "group_ref")} {
+            dict for {iter content} $val {
+                set name $iter
+                set uhdmclasstype uhdm$name
+                if {$key == "group_ref"} {
+                    if ![info exist GROUP_MEMBERS($name)] {
+                        puts "ERROR: Group $name unknown while processing group $groupname"
+                    }
+                    set members $GROUP_MEMBERS($name)
+                    foreach member $members {
+                        set uhdmgroupmember uhdm[lindex $member 0]
+                        append checktype " $uhdmgroupmember"
+                        set member_name [lindex $member 0]
+                        if [info exist GROUP_MEMBERS($member_name)] {
+                            append checktype [recurse_generate_group_checker $NAME_TO_MODEL($member_name)]
+                        }
+                    }
+                } else {
+                    append checktype " $uhdmclasstype"
+                }
+                if {$key == "class_ref"} {
+                    if [info exist ALL_CHILDREN($name)] {
+                        foreach child $ALL_CHILDREN($name) {
+                            set name $child
+                            set uhdmclasstype uhdm$name
+                            append checktype " $uhdmclasstype"
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return $checktype
+}
+
 proc generate_group_checker { model } {
     global $model BASECLASS ALL_CHILDREN GROUP_MEMBERS
     set data [subst $$model]
@@ -710,7 +752,6 @@ proc generate_group_checker { model } {
 
     set files [list [list "[project_path]/templates/group_header.h" "[gen_header_dir]/${groupname}.h"] \
                    [list "[project_path]/templates/group_header.cpp" "[gen_src_dir]/${groupname}.cpp"]]
-    set GROUP_MEMBERS($groupname) ""
     foreach pair $files {
         foreach {input output} $pair {}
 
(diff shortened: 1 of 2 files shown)


hs-apotell:
@alaindargelas Which specific group model would this change impact? I would need an example to confirm both sides are generating the same.

hs-apotell:
I am working with `named_event_sequence_expr_group` as my test case and the generated output with your change seems wrong. The output is the following -

- distribution
- expr
- named_event
- operation
- sequence_inst

which, incidentally, is no different than what it was before your change. However, `named_event_sequence_expr_group` has a group_ref to `sequence_expr_group` which in turn has a class_ref of `expr`. With `expr` in picture, don't all subclasses of `expr` be included in the check as well? I think the output should be the following -

- array_net
- array_var
- bit_select
- bit_var
- byte_var
- chandle_var
- class_var
- constant
- distribution
- enum_net
- enum_var
- expr
- func_call
- hier_path
- indexed_part_select
- int_var
- integer_net
- integer_var
- let_expr
- logic_net
- logic_var
- long_int_var
- method_func_call
- method_task_call
- named_event
- net
- net_bit
- nets
- operation
- packed_array_net
- packed_array_var
- parameter
- part_select
- real_var
- ref_obj
- ref_var
- sequence_inst
- short_int_var
- short_real_var
- simple_expr
- spec_param
- string_var
- struct_net
- struct_var
- sys_func_call
- sys_task_call
- task_call
- tf_call
- time_net
- time_var
- union_var
- unsupported_expr
- var_bit
- var_select
- variables

Could you please confirm. Thanks!

hs-apotell:
Also, should group checks include other groups. For instance, `operand_group` includes a check for `sequence_expr_group`. Is that legal?

alaindargelas:
@hs-apotell, please check the new implementation, 
- it does now include the expr sub classes,
- it sorts the member alphabetically in the generated code
- it eliminates duplication in the generated code
- it prevents loops while traversing the group membership

Please implement the same in the Python generation, then let's try to switch to the Python generation code officially.

hs-apotell:
@alaindargelas This still doesn't look right to me. Please confirm that the implementation is generating the correct list of checks. Specific example which I was working with - `assertion` and `stmt`

`stmt` has a group_ref to `assertion` and no other group_ref. I expected `stmt` checks to include only `assertion` and not all the obj_ref that `assertion` has since `assertion` logic itself doesn't drill down obj_refs. The checks for `stmt` should have superset of `assertion` with just additional checks to include `scope`, `atomic_stmt` and `expr` (and their corresponding subclasses). But the generated code includes a whole lot more.

Please confirm what you have is the expected/intended behavior.

Potential Issue - The use of GROUP_MEMBERS. Do you really need this? When a group_ref is found while iterating the data, shouldn't you just recurse with the model referred to by the reference?

Instead of this -
```
if {$key == "group_ref"} {

    if [info exist GROUP_MEMBERS($name)] {
        set members $GROUP_MEMBERS($name)
    } else {
        set members ""
    }

    foreach member $members {
        set uhdmgroupmember uhdm[lindex $member 0]
        append checktype " $uhdmgroupmember"
        set member_name [lindex $member 0]
        set member_type [lindex $member 1]
        append checktype [recurse_generate_group_checker $NAME_TO_MODEL($member_name)]
    }
} else {
    append checktype " $uhdmclasstype"
}
```

shouldn't this be the logic -

```
if {$key == "group_ref"} {
    append checktype [recurse_generate_group_checker $NAME_TO_MODEL($name)]
} else {
    append checktype " $uhdmclasstype"
}
```

alaindargelas:
I checked the implementation of stmt.cpp it is exactly what is expected.
This is the intent of this change, group_ref recursively accept all group_ref and their members.
This code is correct:

```
if {$key == "group_ref"} {

    if [info exist GROUP_MEMBERS($name)] {
        set members $GROUP_MEMBERS($name)
    } else {
        set members ""
    }

    foreach member $members {
        set uhdmgroupmember uhdm[lindex $member 0]
        append checktype " $uhdmgroupmember"
        set member_name [lindex $member 0]
        set member_type [lindex $member 1]
        append checktype [recurse_generate_group_checker $NAME_TO_MODEL($member_name)]
    }
} else {
    append checktype " $uhdmclasstype"
}
```

alaindargelas:
Also if a class_ref is referenced, all of its concrete class are accepted by the checker.

hs-apotell:
> group_ref recursively accept all group_ref and their members

In that case, shouldn't stmt/scope be recursed as well because stmt/assertion/sequence_inst is. Seems like the root group_ref is getting special treatment because obj_ref and class_ref at the root level don't get recursed.

hs-apotell:
A good example of the inconsistency. When processing `assertion`, the check doesn't include `named_event` because `assertion`/`sequence_inst` never gets recursed but when `assertion` becomes part of `stmt` as a group_ref, `named_event` is included in the checks because `sequence_inst` obj_ref is recursed. The same obj_ref (`sequence_inst` member of `assertion` in this case) is being treated differently based on whether it is part of reference or a definition.

hs-apotell:
For the `stmt` tree presented below -
```
stmt
  assertion (group_ref)
    property_inst (obj_ref)
      property_expr_named_event_group (group_ref)
        property_expr_group (group_ref)
          case_property (obj_ref)
```

Why is `case_property` not part of checks for `stmt`? Similar issue with `clocked_property` and `multiclock_sequence_expr`.

alaindargelas:
In #594, named_event is no more part of the stmt, thanks for catching this.

Here are the rules for group membership:
- A group_def accepts ad member its obj_ref, class_ref (and it's children), group_ref (and its members).
- A obj_def does not accept any field 
- A class_def does not accept any field

So, for the example above
```
stmt
  assertion (group_ref)
    property_inst (obj_ref)
      property_expr_named_event_group (group_ref)  <<<< No member of property_inst (obj_ref) is accepted, this branch should not be followed.
        property_expr_group (group_ref)
          case_property (obj_ref)
```

hs-apotell:
This is starting to make a lot more sense now. I think you are still missing one line.

```
if {$key == "class_ref"} {
    append checktype " $uhdmclasstype"      <== Shouldn't the class_def itself be included??
    if [info exist ALL_CHILDREN($name)] {
        foreach child $ALL_CHILDREN($name) {
            set name $child
            set uhdmclasstype uhdm$name
            append checktype " $uhdmclasstype"
        }
    }
}
```

hs-apotell:
I have made a few changes #597 to the tcl script to simplify the logic. Please review it. PR also include the update to python scripts.

alaindargelas:
That line is optional since classes cannot be instantiated, so it's up to you if it makes the python script matching.

PR #597, opened 2021-12-28
Transitive Grouping

hs-apotell:
Transitive Grouping

Implement transitive grouping in python and also simplify the logic
in tcl script.

### DIFF
--- a/model_gen/model_gen.tcl
+++ b/model_gen/model_gen.tcl
@@ -703,7 +703,7 @@ proc defineType { def name vpiType } {
 }
 
 proc recurse_generate_group_checker { model } {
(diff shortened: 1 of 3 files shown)


hs-apotell:
@alaindargelas Please review the tcl specific changes before merging.

alaindargelas:
@hs-apotell, please submit a PR where the Python code generation becomes the only generation. Leave the Tcl code for reference, but it should not be called from cmake anymore. We can remove the Surelog matrix CI for Tcl too.

hs-apotell:
@alaindargelas Is this transition only for UHDM or include Surelog as well?

alaindargelas:
For both.

Issue #603, opened 2022-01-04
No implementation of vpi_compare_objects function


RRozak:
`vpi_compare_objects` function is unimplemented.

I need it to check if in the example below `vpiParent` refers to this `var_select` or some other if there would be more.
I tried to compare the names, but this function will make it much simpler and should work in general.

```
      \_var_select: (work@top.a.a), line:3:20, endln:3:29, parent:work@top.a
        |vpiName:a
        |vpiFullName:work@top.a.a
        |vpiIndex:
        \_constant: , line:3:22, endln:3:23, parent:work@top.a.a
          |vpiDecompile:0
          |vpiSize:64
          |UINT:0
          |vpiConstType:9
        |vpiIndex:
        \_part_select: , parent:work@top.a.a
          |vpiParent:
          \_var_select: (work@top.a.a), line:3:20, endln:3:29, parent:work@top.a
          |vpiConstantSelect:1
          |vpiLeftRange:
          \_constant: , line:3:25, endln:3:26
            |vpiDecompile:3
            |vpiSize:64
            |UINT:3
            |vpiConstType:9
          |vpiRightRange:
          \_constant: , line:3:27, endln:3:28
            |vpiDecompile:0
            |vpiSize:64
            |UINT:0
            |vpiConstType:9
```

alaindargelas:
@hs-apotell, that new function like vpi_iterate would benefit from being auto-generated. Can you take on that task?
Since we switched to the Python code generation, I would not know where to start.
Basically for 2 given objects, compare that their trees are isomorphic and values are equal, note, the vpifFullName value should not be checked.

hs-apotell:
@alaindargelas Will look into it shortly.

hs-apotell:
@alaindargelas @RRozak Few questions about the potential use cases and implementation -
* Is this supposed to be a shallow compare or a deep compare?
* Is the expected result a boolean (true or false) or integer (less, equal, greater) or are we interested in more detailed information on the actual differences themselves?

alaindargelas:
It should be a deep compare, only equal not equal.

alaindargelas:
Ordered compare, compare in place the vectors

alaindargelas:
Fixed by #618

PR #618, opened 2022-01-20
Issue #603 - Implement vpi_compare_objects

hs-apotell:
Issue #603 - Implement vpi_compare_objects
  * Introducing BaseClass::Compare
  * Cleanup post py migration


hs-apotell:
@RRozak Could you please evaluate this change? This is just the first pass though. Posting the draft to ensure I am in the right direction. Input welcome!

@alaindargelas Please review the change and provide any feedback.

RRozak:
I used this function in handling of select statements: https://github.com/antmicro/verilator/pull/676
Unfortunately many tests failed.
I prepared the simplest test on which it fails (https://github.com/chipsalliance/UHDM-integration-tests/pull/659):
```systemverilog
module top(output logic[1:0] o);
   logic [4:0][1:0] x = '1;
   assign o = x[0][1:0];
endmodule
```
Here is the uhdm of the `x[0][1:0]`:
```
    \_var_select: (work@top.x), line:3:15, endln:3:24            <------------ A
      |vpiName:x
      |vpiFullName:work@top.x
      |vpiIndex:
      \_constant: , line:3:17, endln:3:18, parent:work@top.x
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
      |vpiIndex:
      \_part_select: , parent:work@top.x
        |vpiParent:
        \_var_select: (work@top.x), line:3:15, endln:3:24      <------------ B
        |vpiConstantSelect:1
        |vpiLeftRange:
        \_constant: , line:3:20, endln:3:21
          |vpiDecompile:1
          |vpiSize:64
          |UINT:1
          |vpiConstType:9
        |vpiRightRange:
        \_constant: , line:3:22, endln:3:23
          |vpiDecompile:0
          |vpiSize:64
          |UINT:0
          |vpiConstType:9
```
vpi_compare_objects function on nodes `A` and `B` should return true, but returns false.

hs-apotell:
Thanks! I will use this test case to investigate what’s going on.

hs-apotell:
@RRozak The specific example you provided above seems to work as expected. It is returning true in my testing. In fact, this specific test case isn't even a "good" because it won't even step into the auto-generated Compare function. It will bail out success immediately because both `object1` and `object2` will be the same pointers.

https://github.com/chipsalliance/UHDM/pull/618/files#diff-390e6d5a810104b2e1fd11172f8127492ae691bd55bcd376cb2444ce90f7fdf9R237

Could you please check your test environment to confirm. Thanks!!

RRozak:
The line you posted returns 0 if the objects are equal.
In standard it is said:
`The function vpi_compare_objects()
compares the objects they refer to. It returns the value 1 if the objects they refer to are the same object);
otherwise it returns the value 0. See 38.3`
I changed this line: https://github.com/hs-apotell/UHDM/pull/1, but it still doesn't work.

RRozak:
I see that in other places the values are also swapped.
When I treat the value returned by the function as negated, it still doesn't work. I published PR (https://github.com/antmicro/verilator/pull/679) to see the results on CI and find the minimal failing test case.

RRozak:
I think that the reason of failures (after inserting negation in function return) can be on Verilator side. It can be caused be the difference in meaning of vpiParent field in vpiBitSelect and other selects. I will investigate it further in the next week.

Could you change the implementation to return the values according to the standard? Maybe there are cases in which the current implementation returns appropriate values and if so, negation of the result also doesn't work.

hs-apotell:
First off, my development and testing environment is on Windows and so Verilator examples don't necessarily help me. If you can provide cross-platform examples of what's failing that would be great.

Secondly, this `It returns the value 1 if the objects they refer to are the same object` is counter to what most, if not all, generic C/Java API would do. I don't want to change that expectation for the `BaseClass::Compare` function since that is doing what's expected of a function named `Compare` i.e. if equal returns 0, if less returns -n and otherwise +n. There is nothing in the implementation of `BaseClass::Compare` that would prevent negating the result at the `vpi_compare_objects` level. I will submit a revision with this change.

Also, in my opinion, the statement in the standard doesn't mean _isomorphic_ comparison. Wouldn't a trivial pointer comparison meet the standard requirements (assuming the input `vpiHandle`s are different but pointing to the same UHDM object)? The only scenario, that I can think of, that might fail with that implementation is if one input was from non-elaborated version and other from elaborated version. But in that case should that even be considered equal?

At last, the only reason the implementation would be incorrect is if it is comparing something that it shouldn't. All comparison checks are based on the data type. If anything it could be returning a positive number instead of negative or vice-versa but not returning zero to declare equal. I have excluded the following fields explicitly from being considered.

If _vpi_ is one of the following -
* vpiParent
* uhdmParentType
* uhdmType
* vpiLineNo
* vpiColumnNo
* vpiEndLineNo
* vpiEndColumnNo
* vpiFile
* vpiFullName
* uhdmId

If _type_ in one of the following -
* value
* delay

Note that `vpiParent` is explicitly excluded in comparison. This is done to avoid walking up the reference tree. If you think there are other fields that should be ignored, please make a note. @alaindargelas You might want to review this list as well and comment if necessary.

hs-apotell:
> I changed this line: hs-apotell#1, but it still doesn't work.

Your edit was still incorrect. It should have been the following -
```
return (object1 == object2) ? 1 : ((object1->Compare(object2) == 0) ? 1 : 0);
```
Because a non-zero (i.e. both positive and negative) will still logically be considered `true`.

RRozak:
> Also, in my opinion, the statement in the standard doesn't mean isomorphic comparison. Wouldn't a trivial pointer comparison meet the standard requirements (assuming the input vpiHandles are different but pointing to the same UHDM object)? The only scenario, that I can think of, that might fail with that implementation is if one input was from non-elaborated version and other from elaborated version. But in that case should that even be considered equal?

I think you may be right. In my opinion it doesn't mean the isomorphic comparison too.
In the following example:
```systemverilog
module top;
   int x;
   assign x = 1; // assignment_1
   assign x = 1; // assignment_2
endmodule
```
I would say that the assignment_1 and assignment_2 are not the same object.
From perspective of current Verilator it doesn't matter.
If @alaindargelas doesn't see any problems (he said previously that it should be the isomorphic comparison), I think that the pointer comparison is better.

I also don't know what to return in case of comparison between handle from elaborated tree with the one from non-elaborated. If they should be treated as the same object, the file positions can be compared, so we still wouldn't need to compare whole trees.

alaindargelas:
The problem with pointer (or ID) comparison is that it forces UHDM to actually give back the same pointer. 
In a lot of cases, during the elaboration process the cloning of sub expressions is necessary, the pointer and the ID is not the same but the expression represents the same "object" from the design. 
If you take the example above and print the IDs (which is equivalent to the pointers) you'll notiice:


```
|vpiElaborated:1
|vpiName:work@top
|uhdmallModules:
\_module: work@top (work@top), id:23 tests/ScratchPad.sv:10:1: , endln:13:10, parent:work@top, parID:22
  |vpiFullName:work@top
  |vpiDefName:work@top
  |vpiNet:
  \_logic_net: (work@top.o), id:25, line:10:30, parent:work@top, parID:23
    |vpiName:o
    |vpiFullName:work@top.o
    |vpiNetType:36
  |vpiNet:
  \_logic_net: (work@top.x), id:27, line:11:21, parent:work@top, parID:23
    |vpiName:x
    |vpiFullName:work@top.x
    |vpiNetType:36
  |vpiPort:
  \_port: (o), id:24, line:10:30, parent:work@top, parID:23
    |vpiName:o
    |vpiDirection:2
    |vpiLowConn:
    \_ref_obj: , id:26
      |vpiActual:
      \_logic_net: (work@top.o), id:25, line:10:30, parent:work@top, parID:23
  |vpiContAssign:
  \_cont_assign: , id:7, line:12:11, endln:12:24, parent:work@top, parID:23
    |vpiRhs:
    \_var_select: (work@top.x), id:1, line:12:15, endln:12:24, parID:7         <<< Non elaborated ID:1
      |vpiName:x
      |vpiFullName:work@top.x
      |vpiIndex:
      \_constant: , id:2, line:12:17, endln:12:18, parent:work@top.x, parID:1
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
      |vpiIndex:
      \_part_select: , id:5, parent:work@top.x, parID:1
        |vpiParent:
        \_var_select: (work@top.x), id:1, line:12:15, endln:12:24, parID:7               <<< Non elaborated ID:1
        |vpiConstantSelect:1
        |vpiLeftRange:
        \_constant: , id:3, line:12:20, endln:12:21
          |vpiDecompile:1
          |vpiSize:64
          |UINT:1
          |vpiConstType:9
        |vpiRightRange:
        \_constant: , id:4, line:12:22, endln:12:23
          |vpiDecompile:0
          |vpiSize:64
          |UINT:0
          |vpiConstType:9
    |vpiLhs:
    \_ref_obj: (work@top.o), id:0, line:12:11, endln:12:12, parID:7
      |vpiName:o
      |vpiFullName:work@top.o
      |vpiActual:
      \_logic_net: (work@top.o), id:12, line:10:30, endln:10:31, parent:work@top, parID:28
        |vpiName:o
        |vpiFullName:work@top.o
        |vpiNetType:36
        |vpiRange:
        \_range: , id:9, line:10:25, endln:10:28
          |vpiLeftRange:
          \_constant: , id:10, line:10:25, endln:10:26, parID:9
            |vpiDecompile:1
            |vpiSize:64
            |UINT:1
            |vpiConstType:9
          |vpiRightRange:
          \_constant: , id:11, line:10:27, endln:10:28, parID:9
            |vpiDecompile:0
            |vpiSize:64
            |UINT:0
            |vpiConstType:9
|uhdmtopModules:
\_module: work@top (work@top), id:28 tests/ScratchPad.sv:10:1: , endln:13:10
  |vpiName:work@top
  |vpiVariables:
  \_logic_var: (work@top.x), id:20, line:11:21, endln:11:22, parent:work@top, parID:28
    |vpiName:x
    |vpiFullName:work@top.x
    |vpiAutomatic:1
    |vpiVisibility:1
    |vpiExpr:
    \_constant: , id:19, line:11:25, endln:11:27
      |vpiDecompile:'1
      |vpiSize:-1
      |BIN:1
      |vpiConstType:3
    |vpiRange:
    \_range: , id:13, line:11:11, endln:11:14
      |vpiLeftRange:
      \_constant: , id:14, line:11:11, endln:11:12, parID:13
        |vpiDecompile:4
        |vpiSize:64
        |UINT:4
        |vpiConstType:9
      |vpiRightRange:
      \_constant: , id:15, line:11:13, endln:11:14, parID:13
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
    |vpiRange:
    \_range: , id:16, line:11:16, endln:11:19
      |vpiLeftRange:
      \_constant: , id:17, line:11:16, endln:11:17, parID:16
        |vpiDecompile:1
        |vpiSize:64
        |UINT:1
        |vpiConstType:9
      |vpiRightRange:
      \_constant: , id:18, line:11:18, endln:11:19, parID:16
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
  |vpiDefName:work@top
  |vpiNet:
  \_logic_net: (work@top.o), id:12, line:10:30, endln:10:31, parent:work@top, parID:28
    |vpiName:o
    |vpiFullName:work@top.o
    |vpiNetType:36
    |vpiRange:
    \_range: , id:9, line:10:25, endln:10:28
      |vpiLeftRange:
      \_constant: , id:10, line:10:25, endln:10:26, parID:9
        |vpiDecompile:1
        |vpiSize:64
        |UINT:1
        |vpiConstType:9
      |vpiRightRange:
      \_constant: , id:11, line:10:27, endln:10:28, parID:9
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
  |vpiPort:
  \_port: (o), id:29, line:10:30, endln:10:31, parent:work@top, parID:28
    |vpiName:o
    |vpiDirection:2
    |vpiLowConn:
    \_ref_obj: (work@top.o), id:30, parent:o, parID:29
      |vpiName:o
      |vpiFullName:work@top.o
      |vpiActual:
      \_logic_net: (work@top.o), id:12, line:10:30, endln:10:31, parent:work@top, parID:28
    |vpiInstance:
    \_module: work@top (work@top), id:28 tests/ScratchPad.sv:10:1: , endln:13:10
  |vpiContAssign:
  \_cont_assign: , id:31, line:12:11, endln:12:24, parent:work@top, parID:28
    |vpiRhs:
    \_var_select: (work@top.x), id:33, line:12:15, endln:12:24, parID:31              <<< During elaboration new ID:33
      |vpiName:x
      |vpiFullName:work@top.x
      |vpiIndex:
      \_constant: , id:2, line:12:17, endln:12:18, parent:work@top.x, parID:1
        |vpiDecompile:0
        |vpiSize:64
        |UINT:0
        |vpiConstType:9
      |vpiIndex:
      \_part_select: , id:34, parent:work@top.x.x, parID:35                      <<< During elaboration new ID:34
        |vpiParent:
        \_ref_obj: (work@top.x.x), id:35, parent:work@top.x, parID:1       <<< Non elaborated ID:1 as this pointer does not go through cloning. 
        |vpiConstantSelect:1
        |vpiLeftRange:
        \_constant: , id:3, line:12:20, endln:12:21
          |vpiDecompile:1
          |vpiSize:64
          |UINT:1
          |vpiConstType:9
        |vpiRightRange:
        \_constant: , id:4, line:12:22, endln:12:23
          |vpiDecompile:0
          |vpiSize:64
          |UINT:0
          |vpiConstType:9
    |vpiLhs:
    \_ref_obj: (work@top.o), id:32, line:12:11, endln:12:12, parID:31
      |vpiName:o
      |vpiFullName:work@top.o
      |vpiActual:
      \_logic_net: (work@top.o), id:12, line:10:30, endln:10:31, parent:work@top, parID:28
```

Maybe the solution is to repair the pointer relation after the UHDM elaboration and let be the vpi_compare_object compare pointers or IDs.

hs-apotell:
@alaindargelas So how do we want to move forward on this? Do you want to update the elaboration process to guarantee same IDs or use the isomorphic tree comparison logic?

On a second thought, if there are different UHDM objects representing the same object how are we guaranteeing that they are exactly the same? Isn't it possible that one object gets updated but other doesn't?

@RRozak Did you get a chance to verify your test cases after my last update to fix the `vpi_compare_objects` implementation? Could you please take a moment to report back the result. Thanks!

RRozak:
I tested it before (I negated the results of vpi_comapre_objects function) and many tests failed, but I think that it can be the problem with vpiParents handling in Verilator. I haven't fixed it yet, but I will try in the next days.

hs-apotell:
@RRozak But the implementation you shared in the PR was wrong. Check my previous comment [here](https://github.com/chipsalliance/UHDM/pull/618#issuecomment-1018821326). Since `vpiParent` isn't even traversed as part of comparison implementation makes me wonder how that plays a role.

RRozak:
I talk about this comment: https://github.com/chipsalliance/UHDM/pull/618#issuecomment-1018456279.
I changed there the line to `return not((object1 == object2) ? 0 : object1->Compare(object2));` which does exactly the same as your change.

> Since vpiParent isn't even traversed as part of comparison implementation makes me wonder how that plays a role.
I want to call vpi_compare_objects on vpiParent and some other object. I suspect that there is a bug related to vpiParent handling by Verilator, no to your implementation.

hs-apotell:
> I talk about this comment: #618 (comment).

Sorry, I missed the `not(...)` wrapper around the whole condition.

alaindargelas:
Regarding: "@alaindargelas So how do we want to move forward on this? Do you want to update the elaboration process to guarantee same IDs or use the isomorphic tree comparison logic?"

I think it is really had to predict upfront if an expression should be or not be cloned during elaboration because parameterized expressions. That is why the clone function is brute force and clones everything. So keeping the same pointers is going to be hard. 

On the other hand, the isomorphic check seems also brute force and should work for the general case.
Post elaboration, if expressions are isomorphic they point to the same "object".
If parameters changed one of the expressions, they won't be isomorphic. Which is what we want.

RRozak:
I used your implementation in select handling. All tests passed. I also tested it on this test: https://github.com/chipsalliance/UHDM-integration-tests/blob/11f6ef2492bf07fb0360329278006e95537608ca/tests/SelfSelectsInBitSelectAfterBitSelect/top.sv which fails with current uhdm-verilator. It passes when `vpi_compare_objects` is used.

alaindargelas:
So, can we merge this?

hs-apotell:
@RRozak Appreciate the confirmation.

RRozak:
Yes, in my opinion you can merge.
Thank you for the implementation and discussion.

Issue #640, opened 2022-02-11
Check for relevant python dependency (orderedmultidict) in cmake


hzeller:
Cmake should check that the needed Python `orderedmultidict` library is available.

hzeller:
@hs-apotell can you add a test in cmake for this to improve the install experience ?

hzeller:
Also see https://github.com/chipsalliance/Surelog/issues/2576 for suggestions by @mithro how to add this to cmake.

PR #645, opened 2022-02-13
chipsalliance/UHDM#640: Add a check to find required python packages

hs-apotell:
chipsalliance/UHDM#640: Add a check to find required python packages

Added a check to find 'orderedmultidict' and repport an error if the
module isn't found.


mithro:
`s/pythin/python/`

PR #646, opened 2022-02-13
Reduce template specialization

hs-apotell:
Rework serialization to reduce template specialization and thus reduce generated binary size.
For release builds, this change reduces the binary sizes by ~20%.

Instead of dumping every field in the capnp schema for each class, subclasses carry only a pointer to the base schema. Thus, a class, say `attribute`, always loads data from a schema named `Attribute` and so no template specialization is required.

@alaindargelas @hzeller This is only a prototype of the change and for either of you to play around with if you find this worthwhile. If the change passes any/all checks, then I can clean up and create a PR for merge.

### DIFF
--- a/scripts/capnp.py
+++ b/scripts/capnp.py
@@ -29,10 +29,21 @@ def _get_schema(type, vpi, card):
 
 
 def generate(models):
-    individual_model_schemas = {}
-    flattened_model_schemas = []
     root_schema_index = 2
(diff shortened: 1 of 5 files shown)


alaindargelas:
I don't have time to try this, simply make sure you run a local build with Surelog and run the regression, if it does not blow up runtime or memory, then it seems good to go.

PR #650, opened 2022-02-27
allPackages vs topPackages

alaindargelas:
Introduce allPackages that contain param_assign without constant pushing vs topPackages that contain all objects in packages including param_assign with constant pushing


alaindargelas:
@hs-apotell, you'll find all the objects defined inside a package in the topPackages, moreover you'll find the non-elaborated param_assign definitions in the allPackages

PR #658, opened 2022-03-06
Remove listen_/enter/leave for class_def

hs-apotell:
Remove listen_/enter/leave for class_def

Remove public API and enter/leave listener callbacks for class_def
models.

These methods give the false impression that those callbacks will
be called while walking the design hierarchy. The callbacks were
meant to be called iff the specific listen_<class_def> API was
invoked. Not worth the ambiguity.

P.S. This has infact caused a bug in ElaboratorListener & SynthSubset
implementations where enter/leave functions were implemented for
class_def but those functions will never be called.

Also, fixed a few build warnings in ExprEval

P.S. This is a potentially breaking change since it modifies the public
VpiListener.h file and removes a few publicly available listen_xxx
APIs.

### DIFF
--- a/scripts/VpiListener_h.py
+++ b/scripts/VpiListener_h.py
@@ -5,12 +5,12 @@ import file_utils
 def get_methods(models, enter, leave):
     methods = []
     for model in models.values():
-        if model['type'] != 'group_def':
+        if model['type'] not in ['class_def', 'group_def']:
             classname = model['name']
             Classname_ = classname[:1].upper() + classname[1:]
 
-            methods.append(f'    virtual void enter{Classname_}(const {classname}* object, const BaseClass* parent, vpiHandle handle, vpiHandle parentHandle) {{{enter}}}')
-            methods.append(f'    virtual void leave{Classname_}(const {classname}* object, const BaseClass* parent, vpiHandle handle, vpiHandle parentHandle) {{{leave}}}')
+            methods.append(f'  virtual void enter{Classname_}(const {classname}* object, const BaseClass* parent, vpiHandle handle, vpiHandle parentHandle) {{{enter}}}')
+            methods.append(f'  virtual void leave{Classname_}(const {classname}* object, const BaseClass* parent, vpiHandle handle, vpiHandle parentHandle) {{{leave}}}')
             methods.append('')
 
     return methods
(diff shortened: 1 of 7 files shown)


alaindargelas:
This PR contains way too many things together, it needs to be broken down.

- What do you mean by: " This is a potentially breaking change since it modifies the public
VpiListener.h file and removes a few publicly available listen_xxx
APIs.", 
Have you tried it with Surelog? Does it break Surelog build? Does it create logs difference? Add a log file with the diffs from Surelog if any.

hs-apotell:
> This PR contains way too many things together, it needs to be broken down.

The PR has only two changelist. One which fixes a few compile errors in ExprEval.cpp and other which removes the class_def related calls from visitor and listener. I don't see how I can break this further.

> This is a potentially breaking change

Few functions in vpi_listener.h and vpi_visitor.h are now removed. So, if there are users that are using these functions they are likely to break. Regardless, these are errors so they should fix it either way.

> Have you tried it with Surelog?

Yes. here's the build - https://github.com/hs-apotell/Surelog/actions/runs/1943903812

> Does it break Surelog build? Does it create logs difference?

No and Yes. No build errors but logs do need update. Six test got affected because of the change in ElaboratorListener.cpp. You might want to review the differences in logs to validate them. Note that the implementation in ElaboratorListener::enterVariables & ElaboratorListener::enterTask_func were never getting executed before and now both functions are.

* ArianeElab
* ArianeElab2
* ClassParamAsParam
* ClassTypeParam
* DollarRoot
* FuncDef2

> Add a log file with the diffs from Surelog if any.

The logs are available for download from the build link provided above.

alaindargelas:
You did not answer above why did you remove
SynthSubset::leaveConstraint_expr
And other non synthesizable constructs?

Also why the method you removed only apply to class? I don't understand. Did you mean to remove all non concrete methods like leave_expr (virtual class) and keep only concrete class methods like leave_expr

alaindargelas:
Got it! Thanks

hs-apotell:
@alaindargelas You might want to take a closer look at the implementation for ElaboratorListener::enterVariables just to make sure the logic is still valid for all the changes that have gone in between August, 2021 and now. I haven't noticed any major impact on the regression tests itself except for the 6 noted above. I would recommend reviewing the changes in logs to make sure those are expected result of the new code path introduced.

PR #677, opened 2022-03-14
UhdmListener - Uhdm alias of VpiListener/vpi_listener logic

hs-apotell:
UhdmListener - Uhdm alias of VpiListener/vpi_listener logic

Introducing UhdmListener, logic parallel to what
VpiListener/vpi_listener implements but with a few additional features.

* Includes visitor & callstack as part of the base class implementation
* Adds new enter/leave functions for vector's as well
* Less memory churn since no handle allocation on every call

Also, includes a new class ScopedVpiHandle to create managed VpiHandle
instances on stack (gets automatically destroyed when it goes out of
scope).

Both the visitor and the callstack logic was getting duplicated in every
sub-class implementation and this implements wraps it all together
nicely. The callstack management is specially tedious because it forces
the subclass to implement every entry/leave function. Small changes in
the design can change the call hierarchy significantly, and the
callstack provides a nice way to know the context in which a entry/leave
function was called.


alaindargelas:
Please add:
- Unit test in the tests/ directory
- Explanation in the README with pointer to the unit test like other features (listener/visitor/elaborator)

PR #681, opened 2022-03-19
vpi_visitor & vpi_listener logic update

hs-apotell:
vpi_visitor & vpi_listener logic update

* vpi_listener
** vpiExtends cannot be safely ignored since there could be only a
   single reference to it. Ignoring would mean there is never a
   entry/leave call for it.
** Avoid entry/leave for vpiProgram, vpiPackage, and vpiUdp when
   referred as a member reference from other objects.

* vpi_visitor
** Show vpiParent in the output for bit_select & part_select
** Never use shallow for ref_obj - without the vpiActual the reference
   itself is pointless
** Similar update for vpiInstance, vpiModule, vpiInterface, vpiUse,
   vpiProgram, vpiClassDefn, vpiPackage, and vpiUdp.
** Attempt to show all the information at the outer level but shallow
   in inner level of the hierarchy (basically, flattenning the tree a
   bit so it's easier to visually parse).

### DIFF
--- a/scripts/classes.py
+++ b/scripts/classes.py
@@ -428,7 +428,7 @@ def _get_GetByVpiType_implementation(model):
         content.append(f'  switch (type) {{')
 
     if modeltype == 'obj_def':
-        content.append( '  case vpiParent: return std::make_tuple(vpiParent_, static_cast<UHDM_OBJECT_TYPE>(0), nullptr);')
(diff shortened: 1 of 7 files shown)


alaindargelas:
@hs-apotell, Can we finish #677 before we merge this one?

hs-apotell:
#677 and this are independent changes. Either way, #677 is updated with requested changes.

PR #723, opened 2022-05-29
Issue chipsalliance/Surelog#2985: Remove 'parent' in uhdm output

hs-apotell:
Issue chipsalliance/Surelog#2985: Remove 'parent' in uhdm output

The information is now available as an independent field in the hierarchy which includes the id information as well.

### DIFF
--- a/templates/vpi_visitor.cpp
+++ b/templates/vpi_visitor.cpp
@@ -576,19 +576,6 @@ void visit_object(vpiHandle obj_h, int indent, const char *relation, VisitedCont
         }
       }
     }
-    if (vpiHandle par = vpi_handle(vpiParent, obj_h)) {
-      if (const char* parentName = vpi_get_str(vpiFullName, par)) {
-        out << ", parent:" << parentName;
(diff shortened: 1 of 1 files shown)


alaindargelas:
This is not desired. There is no parent information anymore with this PR.

alaindargelas:
Screenshot from 2022-05-30 17-53-34

hs-apotell:
That doesn't add up. This PR with chipsalliance/Surelog#2985, here's what the CI shows.

image

Here's the build https://github.com/hs-apotell/Surelog/actions/runs/2403668555

I don't see a build with PR chipsalliance/Surelog#2985 on CI.

alaindargelas:
In the run you point out, it looks like a lot of objects do not have a parent information any more, as was my local run showing after I merged the UHDM PR (The one I reverted).
Please double check, all objects should have a parent info, or at least all the ones that used to have, should still have it.

Do not mix the vpi_ids dump with vpiParent change.

Also, if you need a combined UHDM PR + Surelog PR to achieve the final result, in the Surelog PR , submit the log changes so I can analyze

PR #741, opened 2022-06-08
AllIds in Serializer: use canonical insert; make typedef

hzeller:
o Use canonical map insert/emplace in which we don't find()
  and then do the same thing again on insert, but do it in
  one go.
o Make the map used for the ID storage a typedef to improve
  readability (also makes it easier to replace in one place).


hzeller:
or @alaindargelas whoever is first :)

QuantamHD:
LGTM!

alaindargelas:
I'm down with Covid (Today), expect delays!

PR #743, opened 2022-06-09
chipsalliance/Surelog#2962: Flush often

hs-apotell:
chipsalliance/Surelog#2962: Flush often

With large designs, visitation logic can end up buffering too much with
no progress reported on screen. Switching to flush often. This also
helps with keeping the memory requirement down for large designs.

### DIFF
--- a/templates/vpi_visitor.cpp
+++ b/templates/vpi_visitor.cpp
@@ -577,14 +577,17 @@ void visit_object(vpiHandle obj_h, int indent, const char *relation, VisitedCont
 
 // Public interface
 void visit_designs(const std::vector<vpiHandle>& designs, std::ostream &out) {
+  out << std::unitbuf;
   for (auto design : designs) {
     VisitedContainer visited;
     visit_object(design, 0, "", &visited, out);
   }
+  out << std::nounitbuf;
 }
 
 std::string visit_designs(const std::vector<vpiHandle>& designs) {
   std::stringstream out;
+  out << std::unitbuf;
   visit_designs(designs, out);
   return out.str();
 }


templates/vpi_visitor.cpp
```diff
@@ -577,14 +577,17 @@ void visit_object(vpiHandle obj_h, int indent, const char *relation, VisitedCont
...
   for (auto design : designs) {
     VisitedContainer visited;
     visit_object(design, 0, "", &visited, out);
   }
+  out << std::nounitbuf;
 }
 
 std::string visit_designs(const std::vector<vpiHandle>& designs) {
   std::stringstream out;
+  out << std::unitbuf;
```
hzeller, line 590: This writes into one big string anyway (we accumulate everything that is then returned at the end), so I don't think this will make any difference.

hzeller:
It sounds expensive (lots of system calls) if every write will flush.

Do we know that large allocation is actually an issue here ? I'd suspect the buffer size around a file-stream or `cout` would typically be smallish in the `4k..few-tens-of-k` range.

If too much memory use is a problem from callers using the function returning a string, we might need to solve this by _not_ encouraging doing the wrong thing: If we _remove_ the `std::string visit_designs()` function that returns a string, but rather have the user call `visit_designs(ostream)`, then they have to make the conscious decision to do the expensive thing.

Sometimes convenience functions are just too much rope.

hs-apotell:
I too expected that but, ironically, at least on Windows RSD test runs 15% faster. I checked with both cl and WSL.

This API is not in the critical path for parsing nor for elaboration. It exists for the sake of debugging. I guess a little longer rope is _okay_.

hs-apotell:
@hzeller I can't confirm for sure but it does look like the chipsalliance/Surelog#2962 was this buffer issue. I have got at least 6 consecutive successes and the current master also seems to be succeeding. Unfortunately, it will be hard to confirm conclusively.

As for the buffer size for output streams, there is no standard and platform/compiler dependent. Some compilers tend to auto-flush on every newline and other only on when the buffer is full or other events like cin (which would force a flush on cout). I haven't found any specific details on how gcc or clang works.

hzeller:
This change will call the write() system call for every small thing and it is very, _very_ slow
Now `uhdm-dump` for Rsd alone takes 5 minutes or so...

The main reason things were accumulating memory is, that Surelog actually used the versions that returned std::string.

So instead we should
  1) Undo the this flush-every-time (so that it is fast again)
  2) Patch Surelog to not use the rope and allocate large strings. It is indeed used wrongly in the most expensive way. 
      ( https://github.com/chipsalliance/Surelog/pull/3035 )

PR #744, opened 2022-06-10
Use a std::deque instead of a std::vector to keep track of allocated objects

hzeller:
The FactoryT does not have a need to keep the list contiguously in
memory, so we can use deque, avoiding unnecessary re-allocs.

This allows the memory allocator to not worry about providing many large contiguous regions, which could contribute to an improved situation for https://github.com/chipsalliance/Surelog/issues/2962


hzeller:
@hs-apotell

hs-apotell:
LGTM

PR #745, opened 2022-06-10
Let the stream handle its own flushing. Only use method returning string in tests 

hzeller:
* Changing the flushing on each output creates a huge slowdown
   and leaves the user with no choice. If the caller would like to
   have that behavior, it should be done outside of this function.
 * As looking at Surelog shows, the visitor that returns the string is
   mis-used 100% of the time. Let's not provide that rope.
 * Add the convenience function that retunrs a string in the
   test_utils.h for local use in the unit tests.

Context:
  https://github.com/chipsalliance/UHDM/pull/743
  https://github.com/chipsalliance/Surelog/issues/2962


hzeller:
cc @hs-apotell

PR #746, opened 2022-06-10
Fix more places that overzealously flush the stream.

hzeller:
Using std::endl will flush the stream with each line which is
definitely not what we want when dumping out a large design.

This speeds up uhdm-dump significantly.


hzeller:
verified with `strace`: without this, we get write()-calls for each line, with it the regular buffer size of the std::cout stream, which is 4k in my case (which is a typical size on intel platforms as this is exactly one page).

hzeller:
cc: @hs-apotell 
context: https://github.com/chipsalliance/Surelog/issues/2962
#745 #743

PR #747, opened 2022-06-11
Using std::deque<> as an allocation space for factories.

hzeller:
This allows to have pre-sized buffers and easily de-allocated objects.
Since malloc allocators already do something similar, this is not
creating huge saving (about 70 MiB RAM saving for Rsd, 5-10% CPU time)

This particular implementation will increase the compile time by
a lot as we need to pull in all the classes from uhdm/uhdm.h (as
we need to know the size in the factory), so it would need some
refinement (e.g. some PIMPL-style) to be usable.

So leaving it here as draft if someone wants to experiment with
it.

### DIFF
--- a/templates/BaseClass.h
+++ b/templates/BaseClass.h
@@ -148,16 +148,13 @@ namespace UHDM {
   template<typename T>
   class FactoryT final {
     friend Serializer;
-    // TODO: make this an arena: iinstead of pointers, store the
-    // objects directly with placement-new'ed object using emplace_back()
-    // One less indirection and a lot less overhead.
-    // (dqque guarantees pointer stability; however, we'd need to
-    // do something special in Erase() - can't re-arrange).
+    // TODO: consolidate allocator_ and objects_.
+    // We need both currently  because of Erase()
     typedef std::deque<T *> objects_t;
 
     public:
       T* Make() {
-        T* obj = new T;
+        T* obj = &allocator_.emplace_back();
         objects_.push_back(obj);
         return obj;
       }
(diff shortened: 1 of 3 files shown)


hzeller:
FYI @hs-apotell

hs-apotell:
This problem is actually harder than it seems on the surface. Deleting the objects within `std::deque` is trivial (call the destructor explicitly since the object was created using inplace new) but the memory still can't be claimed or reused safely. There is a likelihood that something or someone might still be holding a pointer to this deleted instance and try to use it. If reclaimed and reused, it will end up doing things to the new instance. Nightmare to debug these issues in large systems. In short, objects can be deleted but the memory can't be reclaimed, making the whole exercise pointless.

In the past, for situation like these, a handle based object allocation with _block allocation scheme_ has worked well. No-one holds a reference to any object. Everybody only has a handle that should be dereferenced/resolved within scope and then discarded. There are multiple O(1) handle based system strategies that fit well for problem like these. These systems also easily lend themselves to garbage collection and memory packing strategies on top. This is, however, a big change to the way things are done today.

All that said, by the time Surelog is _feature complete_, all compilers _should_ support enough features of std::pmr to not have to worry about these problems. The std::pmr module has a lot to offer and as long as used wisely, these memory related performance issues can be stories of the past (fingers crossed). Let's not try and reinvent the wheel, once again!

hzeller:
Well, if someone is keeping a reference or using the memory beyond the lifetime of factories, this is wrong code now and then.

Main issue is that Windows has a brain-dead std::deque implementation that allocates each element individually instead of using a buffer, so it won't help there and possibly even create overhead.

But indeed, there are other fish to fry first.

hzeller:
No, the lifetime implications are the same. It will work _exactly_ as the current factory, as the lifetime are exactly the same as the original.

But it is not necessarily too beneficial at this point as it saves only a little memory and CPU.

PR #748, opened 2022-06-11
Serializer: always Purge() on deconstruct.

hzeller:


hzeller:
(Tested also with Surelog and works there)

PR #790, opened 2022-09-02
Adding support for interface_array & module_array in non-elab model

hs-apotell:
Adding support for interface_array & module_array in non-elab model

### DIFF
--- a/include/sv_vpi_user.h
+++ b/include/sv_vpi_user.h
@@ -88,6 +88,7 @@ extern "C" {
 #define vpiSequenceTypespec                   696
 #define vpiPropertyTypespec                   697
 #define vpiEventTypespec                      698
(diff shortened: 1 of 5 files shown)


include/sv_vpi_user.h
```diff
@@ -88,6 +88,7 @@ extern "C" {
 #define vpiSequenceTypespec                   696
 #define vpiPropertyTypespec                   697
 #define vpiEventTypespec                      698
+#define vpiModuleTypespec                     768
```
alaindargelas, line 91: A change to the standard file has to be clearly indicated with a comment

alaindargelas (commented):
In Surelog this create arrays in the elaborated model unlike what is stated in the title here.
Which is not good.

PR #799, opened 2022-10-01
Code maintenance

hs-apotell:
Code maintenance

* Don't clone strings where not necessary
* Use nullptr instead of 0
* Use std::string::find instead of strstr
* When looking for type prefixes, look for prefix instead of anywhere in the string
* Avoid iterations where unnecessary
* Use any_cast instead of dynamic_cast

### DIFF
--- a/templates/ExprEval.cpp
+++ b/templates/ExprEval.cpp
@@ -233,7 +233,7 @@ any *ExprEval::getValue(const std::string &name, const any *inst,
         operation *op = (operation *)result;
         UHDM::ExprEval eval;
         expr *res = eval.flattenPatternAssignments(s, op->Typespec(),
-                                                  (UHDM::expr *)result);
+                                                   (UHDM::expr *)result);
         if (res->UhdmType() == uhdmoperation) {
           ((operation *)result)->Operands(((operation *)res)->Operands());
         }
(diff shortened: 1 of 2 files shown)


alaindargelas:
@hs-apotell  Please fix the conflict.

templates/ExprEval.cpp
```diff
@@ -3145,25 +3129,21 @@ expr *ExprEval::reduceExpr(const any *result, bool &invalidValue,
...
                 uint64_t cast_to = 0;
-                if (itps->VpiValue().empty()) {
+                if (val.empty()) {
                   cast_to = 32;
+                } else if (val.find("UINT:") == 0) {
+                  cast_to = std::strtoull(
+                      val.c_str() + std::string_view("UINT:").length(), nullptr,
+                      10);
                 } else {
-                  if (itps->VpiValue().find("UINT:") != std::string::npos) {
```
alaindargelas, line 3152: That is an incorrect change

alaindargelas (changes requested):

templates/ExprEval.cpp
alaindargelas: You are treating UINT and INT the same

alaindargelas: OK, I got confused by the diff.

PR #816, opened 2022-11-08
Switch to current googletest release.

hzeller:
This was accidentally undone in
46e60af9af2255650558bab3ee562a5c027618dc


hzeller:
Once merged and pulled to your local client - if you then do `git submodule update --init --recursive` on your machine, it is less likely to be reverted back.

We need the change to 1.12.x as the `testing::TempDir()` was not configurable via environment variables  in Darwin.

PR #845, opened 2022-12-23
Issue #831: Fix compile issues on MinGW

hs-apotell:
Issue #831: Fix compile issues on MinGW

* Use WIN32_LEAN_AND_MEAN when building on Windows to avoid pulling in irrelevant headers
* Rename "interface" to "interface_inst" to avoid conflicting with system defined class with the same name.
* Rename "module" to "module_inst" since the latter is now a C++ reserved word as of C++ 20.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,10 @@ else()
   )
 endif()
 
+if (WIN32)
+  add_compile_definitions(WIN32_LEAN_AND_MEAN)
+endif()
+
 # All the files the generator depends on.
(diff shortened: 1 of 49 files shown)


alaindargelas:
@hs-apotell, I'm just realizing now what changed here.
There was no reason to change vpiModule into vpiModuleInst. Same with vpiInstance,. These are the STANDARD Verilog object model names. They pose no issue to the MinGW compiler.... 
Please submit a PR ASAP to revert back to vpiModule and vpiInstance.
The "module" class can be changed to module_inst, but not the vpiModule name which is in the standard.

alaindargelas:
I opened https://github.com/chipsalliance/UHDM/issues/859

PR #854, opened 2023-01-11
Use ComponentType::value_map instead of std::make_pair

hzeller:
This improves compatibility in case the map type is not std::map.


hzeller:
That should work as well.

PR #874, opened 2023-03-05
Fix system-provided `gtest` use

timkpaine:
Sorry for push noise, unfortunately there was a lot of annoying GitHub actions nuances to look at

This PR does a few things:
- fixes a small non-typo "typo" introduced in https://github.com/chipsalliance/UHDM/pull/817. Both [`FindGTest`](https://cmake.org/cmake/help/latest/module/FindGTest.html) and the local `CMakeLists.txt` file reference `GTEST_INCLUDE_DIRS`, but the code used `GTEST_INCLUDE_DIR` (gosh don't we all love CMake 😆). This is allegedly "private" in `FindGTest`.
- Splits the `ubuntu` test to test both vendored-dependency and non-vendored-dependency code paths for cmake (`ubuntu` chosen because it's usually the fastest on GitHub Actions). Can see behavior on my [fork actions](https://github.com/timkpaine/UHDM/actions)
- fixes a codepath in `CMakeLists.txt` that only linked when not vendoring `capnproto`

---

A few things discovered along the way:
- You need to install googletest from source on Mac to avoid ugliness:
    - https://gitlab.kitware.com/cmake/cmake/-/issues/22238
    - [This line](https://github.com/chipsalliance/UHDM/blob/a7945203690ee595924207308a835fb5a493eb47/tests/symbol_factory_test.cpp#L64) in `symbol_factory_test.cpp` requires C++17-compiled `gtest`, https://github.com/google/googletest/issues/2021

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,12 @@ jobs:
   linux-gcc:
     name: ubuntu-gcc
     runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        vendored_dependencies: [true, false]
+
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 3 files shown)


timkpaine:
Tests look good (finally)

Screenshot 2023-03-05 at 18 20 20

PR #879, opened 2023-03-19
CMake fixes for UHDM used as a vendored git submodule

timkpaine:
A handful of `CMAKE_` variables have to become `PROJECT_` variables so that `UHDM` will use its own paths when vendored as a submodule of `Surelog`


timkpaine:
@alaindargelas should fix the issues on Surelog

PR #880, opened 2023-03-20
Tag on merge whenever version in cmake changes.

hzeller:


hzeller:
Do you know if this will work @timkpaine ? I am not sure if the CI action has permission to do the tagging or if there would be some github action magic needed.

timkpaine:
lgtm, I think whatever allows us to (safely) eliminate my silly `git` checks is best

PR #881, opened 2023-03-22
pkg-config: base inc dir; choose libs depending on USE_HOST_CAPNP

hzeller:
In all our examples, we use `#include <uhdm/foo.h>`, so the base include directory should be $PREFIX/include, not $PREFIX/include/uhdm.

The libraries to link depends on of iwe used USE_HOST_CAPNP. If we compiled and installed from third_party/, we need to include the linking of the libraries from there, otherwise we need to require to use the pkg-config from capnp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -384,6 +384,13 @@ configure_package_config_file(
   NO_SET_AND_CHECK_MACRO
   NO_CHECK_REQUIRED_COMPONENTS_MACRO)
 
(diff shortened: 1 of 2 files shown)


hzeller:
CC: @timkpaine does this look reasonable ? I was not sure if there is a conditional variable expansion inside the file generate command, so I set variables manually.

timkpaine:
Lgtm

PR #886, opened 2023-03-22
Always extract version number from cmake project

hzeller:
Only use git to extract a commit hash if available.


hzeller:
CC: @timkpaine

Issue #888, opened 2023-03-22
Need to run git tagging in CI action - permission denied


hzeller:
I try to [tag the revision whose SoT is in the CMake file](https://github.com/chipsalliance/UHDM/blob/master/.github/workflows/main.yml#L266-L293) on push whenever it changes.

However, it looks like we need some additional permissions in the github action.
https://github.com/chipsalliance/UHDM/actions/runs/4495222394/jobs/7908625709

I don't have access to the 'settings' tab in UHDM, which I suspect I need to find the right settings.

Tagging a few people who might know how to achieve this
@mithro @kgugala @umarcor

kgugala:
#889 attempts to fix this

hzeller:
Looks like we need a bot account; @kgugala fills in details.

For now, I've just manually tagged the release, so that we can explore that permission issue separately.

PR #890, opened 2023-03-24
Set permissions to write tag.

hzeller:
Fixes: #888


hzeller:
Does this look right, @proppy ?

proppy:
Looks good! I didn't realize you could put permission at the job level, handy ;)

proppy (approved):

hzeller:
Well, this was more a question for you, but we'll see how the next tagging works :)

PR #891, opened 2023-03-24
install shared libraries in same folder as static libraries

timkpaine:
xref: https://github.com/chipsalliance/UHDM/pull/891
@hzeller

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -346,6 +346,7 @@ endif()
 install(
   TARGETS uhdm uhdm-cmp uhdm-dump uhdm-hier
   ARCHIVE DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/uhdm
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -346,6 +346,7 @@ endif()
 install(
   TARGETS uhdm uhdm-cmp uhdm-dump uhdm-hier
   ARCHIVE DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/uhdm
+  LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/uhdm
```
hzeller, line 349: While at it, in Surelog, we use `${CMAKE_INSTALL_LIBDIR}/uhdm` (which then ends up installing in lib64 on some systems). Is that the more canonical form we should use here as well ?

timkpaine: yeah we should probably make these the same

PR #892, opened 2023-03-24
Use GNUInstallDirs; put uhdm and capnp into same lib dir.

hzeller:


hzeller:
@timkpaine this unifies the way we do lib directories in Surelog and UHDM.

PR #915, opened 2023-04-13
Expand directory name variables in UHDMConfig.cmake

hzeller:
The library directory might be $PREFIX/lib/uhdm or $PREFIX/lib64/uhdm depending on the system. This takes care of it.


hzeller:
CC: @timkpaine

PR #920, opened 2023-04-16
use more standard cmake args in python folder

timkpaine:


Thomasb81:
I see that but cannot figure out what was legacy from recommended practice.

timkpaine:
`find_package(PythonLibs)` calls [FindPythonLibs](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html) which is deprecated, and for my Ubuntu 22 system didn't yield the right results on latest cmake.

PR #947, opened 2023-05-23
Model class hierarchy change for select nodes

hs-apotell:
Model class hierarchy change for select nodes

bit_select, and other select classes use the vpiParent pointer (as refobj) to provide parent and actual_group access. bit_select.vpiParent.vpiParent is the "true" parent of the bit_select and bit_select.vpiParent.actual_group is access to the actual. However, vpiParent edges in the graph are weak references and ignored for all traversals (including VpiListener, UhdmListener, and vpi_visitor). Tail of parent edges are missing in the UHDM output since the nodes are ignored. The traversal mode also generate unexpected results during runtime because of the ignored edges.

Changing the hierarchy so that ref_obj works as a base (intermediate class in the class hierarchy, yet instantiable) and other select classes (including bit_select, indexed_part_select, part_select, and var_select) are subclasses of ref_obj. This gives access to ref_obj.actual_group and ref.vpiParent can be used for the usual parenting purposes.

* Updated class file generation logic to support generating class files for intermediate nodes in the class hierarchy.
* Updated vpi_visitor, ExprEval, and clone_tree logic to compensate for the model hierarchy change.

* Few code improvements in ElaboratorListener and BaseClass
* Moved UhdmId, VpiFile, and VpiParent to BaseClass since these are required properties for all models in the class hierarchy

### DIFF
--- a/scripts/serializer.py
+++ b/scripts/serializer.py
@@ -10,6 +10,7 @@ def generate(models):
     factory_function_declarations = []
     factory_function_implementations = []
     factory_purge = []
+    factory_gc = []
     factory_stats = []
     factory_object_type_map = []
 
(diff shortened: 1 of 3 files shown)


alaindargelas:
Large amounts of failures in Surelog regression and Yosys plugin code need to be addressed first.

PR #961, opened 2023-06-10
Fixing parenting issues

hs-apotell:
Fixing parenting issues


alaindargelas:
@hs-apotell please resolve issues.

Issue #979, opened 2023-06-29
Regression: `Compare` method of some types can lead to segmentation fault .


mglb:
This change: https://github.com/chipsalliance/UHDM/commit/e2bd08a1ed676ac4dd36706d63c6a2927d61c421#diff-7be1c98e114a3cee6c55636c1ad92f30fca12503978ddbe258b352d7c9469022L545-R545
Makes the comment in the next line not valid in some cases, which can lead to segmentation fault in `Compare` method when `this` and `other` have different types.

# Reproducible example

dut1.sv:
```systemverilog
module dut(input logic [1:0] i, output logic o);
  int n1, n2;
  initial begin
      o = i[0];
  end
endmodule
```

dut2.sv:
```systemverilog
module dut(input logic [1:0] i, output logic o);
  int n1, n2;
  initial begin
      n1 = n2;
  end
endmodule
```

Shell commands:
```sh
# surelog -odir dut1 -parse -elabuhdm dut1.sv
# surelog -odir dut2 -parse -elabuhdm dut2.sv
# uhdm-cmp dut1/slpp_all/surelog.uhdm dut2/slpp_all/surelog.uhdm
Segmentation fault
```

# Explanation

Let's consider calling `bit_select::Compare`, which is the case in the example above. The second type is not important for the purpose of explanation. In the example it is `ref_obj`.
Note that `bit_select` inherits `ref_obj`.

## Generated sources

`bit_select::Compare`:
```c++
int32_t bit_select::Compare(const BaseClass *other, CompareContext* context) const {
  int32_t r = 0;
  if (!context->m_visited.insert(this).second) return r;
  if ((r = basetype_t::Compare(other, context)) != 0) return r;

  const thistype_t *const lhs = this;
  const thistype_t *const rhs = (const thistype_t *)other;

  if ((r = (lhs->VpiConstantSelect() == rhs->VpiConstantSelect()) ? 0 : (lhs->VpiConstantSelect() ? 1 : -1)) != 0) {
    context->m_failedLhs = lhs;
    context->m_failedRhs = rhs;
    return r;
  }

  // ...
}
```

`ref_obj::Compare`:
```c++
int32_t ref_obj::Compare(const BaseClass *other, CompareContext* context) const {
  int32_t r = 0;
  if (!context->m_visited.insert(this).second) return r;
  if ((r = basetype_t::Compare(other, context)) != 0) return r;

  // ...
}
```

## Execution flow

[...]

PR #980, opened 2023-06-30
Issue #979: Casting to wrong type causes crash in Compare function

hs-apotell:
Issue #979: Casting to wrong type causes crash in Compare function

Added a check in all obj_def implementation to check the corresponding object type before casting.


mglb:
Works for me, solves issue reported in #979.

PR #984, opened 2023-07-12
Introducing ref_typespec, parallel to ref_obj for referencing typespecs

hs-apotell:
Introducing ref_typespec, parallel to ref_obj for referencing typespecs

scope::Typespecs are cloned from non-elaborated tree to the elaborated tree.
However, at the moment, this collection doesn't have all the typespecs in the
non-elaborated tree. 

This solves a few different problems -

* Avoids too many duplicates in the elaborated tree.
* For functions like $size and $bits, the parameter binding doesn't work
  because the parameter is a type and not an object. ref_typespec solves
  that binding problem.
* ref_typespec will hold the type location information in parameter
  declaration.

Known Issues -
* Need to collect all typespecs in the scope subtree to populate
  scope::Typespecs
* Though the typespecs are cloned, elaborated tree is still using the ones
  from the non-elaborated tree. The tree is still complete but typespecs are
  crossing the boundary between elaborated and non-elaborated. Need to fix
  this during binding.

### DIFF
--- a/model/packed_array_typespec.yaml
+++ b/model/packed_array_typespec.yaml
@@ -36,18 +36,18 @@
     vpi: vpiRightRange
     type: expr
     card: 1
-  - group_ref: elem_typespec
+  - class_ref: elem_typespec
     name: elem typespec
     vpi: vpiElemTypespec
-    type: enum_struct_union_packed_array_typespec_group
+    type: ref_typespec
     card: 1
   - class_ref: typespec
     name: typespec
     vpi: vpiIndexTypespec
-    type: typespec
+    type: ref_typespec
     card: 1
   - obj_ref: resolution_func
     name: resolution func
     vpi: vpiFunction
     type: function
-    card: 1    
+    card: 1
--- a/templates/ExprEval.cpp
+++ b/templates/ExprEval.cpp
@@ -322,11 +322,13 @@ any *ExprEval::getValue(std::string_view name, const any *inst,
     }
     if (result && (result->UhdmType() == uhdmoperation)) {
       operation *op = (operation *)result;
-      ExprEval eval;
-      if (expr *res = eval.flattenPatternAssignments(s, op->Typespec(),
-                                                     (expr *)result)) {
-        if (res->UhdmType() == uhdmoperation) {
-          ((operation *)result)->Operands(((operation *)res)->Operands());
+      if (const ref_typespec *rt = op->Typespec()) {
+        ExprEval eval;
+        if (expr *res = eval.flattenPatternAssignments(s, rt->Actual_group(),
+                                                       (expr *)result)) {
+          if (res->UhdmType() == uhdmoperation) {
+            ((operation *)result)->Operands(((operation *)res)->Operands());
+          }
         }
       }
     }
@@ -421,8 +423,10 @@ any *ExprEval::getObject(std::string_view name, const any *inst,
 
         const class_defn *base_defn = nullptr;
         if (const extends *ext = defn->Extends()) {
-          if (const class_typespec *tp = ext->Class_typespec()) {
-            base_defn = tp->Class_defn();
+          if (const ref_typespec *rt = ext->Class_typespec()) {
+            if (const class_typespec *tp = rt->Actual_group<class_typespec>()) {
+              base_defn = tp->Class_defn();
+            }
           }
         }
         defn = base_defn;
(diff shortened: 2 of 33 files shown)


Thomasb81:
VPI API is part of the Ieee1800 standard.

I assume that an application using the VPI API is supposed to be compatible with other vendor that implement this API.
It is not clear to me if such change remain compatible, and if it is acceptable. 

In particular I see : 
* `vpiRefTypespec`  code in sv_vpi_user.h but there is already `vpiModuleTypespec` which create an antecedent ...
* at least one vpi attribute is removed from yaml data model `vpiClassTypespec` in class_defn.yaml

Thomasb81:
I do not pretend to fully understand the SV data model and it's yaml representation. So my comment on this subject may be irrelevant.

> `vpiRefTypespec` is a reference to a `typespec` and `vpiModuleTypespec` is the `typespec` itself. I am not seeing the problem here. This is very similar to what `ref_obj` does. `ref_obj` is the reference that can refer to other object types. Similarly, `ref_typespec` is a reference that can point to any other `typespec`.

`ref_obj` is explicitly part of the API (see chapter 37.15), while this introduction is of ref_typespec bring probably some advantage but is not mention in chapter 37. So from my opinion if there is benefice to use reference, it should probably occurred behind the scene of what is reflected by the uhdm db print.

Thomasb81:
> (Unsure what the difference is with dotted line vs. solid line in the diagram)

My naive interpretation of 37.4 is ref_obj inherit of typespec definition (class reference) while it has an attribute ref_obj (Object reference) in other words a pointer allowing to identify parent object.

So typespec should probably be an abstract class (in a C++ implementation) : all other typespec should inherit of it.

Thomasb81:
> Also, the standard uses `type_reference` in a few places but doesn't actually define what it means.

From what I see, this term in not used in the data model chapter. It is used in the grammar description.

alaindargelas:
@hs-apotell, please use ref_obj as it is already massively use as a handle to any sorts of objects.
Please note that we cannot merge this in for now. the yosys_plugin has not absorbed the latest disruptive change you submitted, but that should happen in the next few days.
But then, it would be best if you can submit a PR also in the yosys-plugin to adjust the code to this change.

model/class_defn.yaml
```diff
@@ -46,8 +46,3 @@
     vpi: vpiDerivedClasses
     type: class_defn
     card: any
-  - obj_ref: class_typespecs
```
alaindargelas, line 49: Keep this relation, it is on page 1007 of the IEEE Standard 1800-2017

model/packed_array_typespec.yaml
```diff
@@ -36,18 +36,18 @@
     vpi: vpiRightRange
     type: expr
     card: 1
-  - group_ref: elem_typespec
+  - class_ref: elem_typespec
     name: elem typespec
     vpi: vpiElemTypespec
-    type: enum_struct_union_packed_array_typespec_group
```
alaindargelas, line 42: We become too permissive here. This group (like other groups) it meant to provide a runtime check that we don't construct an illegal model by checking that only the subset of classes are allowed to be used in this relation.

scripts/ElaboratorListener_cpp.py
```diff
@@ -85,35 +85,20 @@ def _generate_module_listeners(models, classname):
...
           listeners.append( '      clone_vec->push_back(obj);')
           listeners.append( '    }')
           listeners.append( '  }')
           listeners.append( '}')
 
-        elif method in ['Ref_modules', 'Gen_stmts']:
-          pass # No elab
-
-        elif method not in ['Ports', 'Nets', 'Parameters', 'Param_assigns', 'Interface_arrays', 'Module_arrays']:
+        elif method not in ['Nets', 'Parameters', 'Param_assigns', 'Interface_arrays', 'Module_arrays']:
```
alaindargelas, line 95: Can we leave out this change?
It should be in a different PR, I'm not sure it is even correct.

scripts/ElaboratorListener_cpp.py
```diff
@@ -85,35 +85,20 @@ def _generate_module_listeners(models, classname):
...
           listeners.append(f'if (auto vec = defMod->{method}()) {{')
           listeners.append(f'  auto clone_vec = serializer_->Make{Cast}Vec();')
           listeners.append(f'  inst->{method}(clone_vec);')
           listeners.append( '  for (auto obj : *vec) {')
-          listeners.append( '    auto* stmt = obj->DeepClone(inst, context_);')
-          listeners.append( '    clone_vec->push_back(stmt);')
-          listeners.append( '  }')
-          listeners.append( '}')
-
-        elif method in ['Ports']:
```
alaindargelas, line 110: What is happening to Ports here? Please don't change this in this PR

templates/ExprEval.cpp
```diff
@@ -39,8 +39,7 @@
 #include <sstream>
 #include <string_view>
 
-using namespace UHDM;
-
```
alaindargelas, line 43: Most changes in this file are irrelevant to the problem at hand.
Do not perform these changes in this PR.

PR #985, opened 2023-07-14
Fixes, Improvements and Code Maintenance. Also, implement GC

hs-apotell:
Fixed, Improvements and Code Maintenance. Also, implement GC

* UhdmListener -Remove duplicate calls to listen VpiParent
* Comparison logic - Treat empty containers and null pointers the same. Moved the comparison logic out from script into BaseClass
* Serialization scripts - Generate output in predictable order so the generated files are comparable
* Code Maintenance - Fixed number of warnings reported in Windows build (mostly local variable hiding previously declarred variable of the same name)
* Uhdm DB
  * Reduce index size from 64 to 32 bits
  * Added versioning
  * Fixed another bug where the expectation that UhdmId are unique was violated. Highest `objId` needed to be preserved across load/save of the db.
  * Similar issue was found with `allIds` bookkeeping. `allIds` is only valid during save operation but the logic around the code assumes that this container is accumulation of all objects.
* Finally, implement garbage collection
  * Replaced n*2 algorithm with a linear to delete
  * Moved certain temporary members that were valid only during load/save operation to be local.


alaindargelas:
@hs-apotell, Can you fix the Python build? Some header is missing it looks like after your refactoring.
https://github.com/chipsalliance/UHDM/actions/runs/5559067810/jobs/10154812044?pr=985

PR #986, opened 2023-07-18
Upgrade cap'n'proto to head

hs-apotell:
Upgrade cap'n'proto to head


hzeller:
What is happening [here in the windows compile](https://github.com/chipsalliance/UHDM/actions/runs/5590887304/jobs/10221219293?pr=986#step:6:303) ?

hzeller:
Maybe you need to do `git submodule update --init --recursive` for your version of UHDM being updated @alaindargelas ?

PR #999, opened 2023-08-08
Lock capnproto to v1.0.0

hs-apotell:
Lock capnproto to v1.0.0

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,9 +62,10 @@ jobs:
 
     - name: Install vendored dependencies
       run: |
-        git clone https://github.com/google/googletest.git
+        git clone --depth 1 https://github.com/google/googletest.git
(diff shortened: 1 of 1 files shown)


timkpaine (approved):
lgtm, idk if you want to just lock to v1 for CI/CD or if you want to force use of v1 (in which case we should add to here: https://github.com/chipsalliance/UHDM/blob/2a9475691e0d48fe7669e8ed77c942eda10b7e05/CMakeLists.txt#L58 )

hzeller:
is this needed ? I found that capnproto even down to 0.7 works just fine.
Requiring vert specific versions makes things hard to package.

hzeller:
> The reason to lock is that the tip now needs gcc 11 or higher because of C++ 20 requirements. Surelog is still on gcc 9. Capnproto v1.0 is the last good release which still works with gcc 9.

Sounds good.

PR #1002, opened 2023-08-17
Remove uhdm libs out of own folder, use soversion (fixes #992)

### LINKED ISSUE #992: libuhdm.so should have a numbered SONAME
Currently the SONAME is ```libuhdm.so```.

But you make incompatible API changes. For example, ```BaseClass::Compare()``` got a new argument in the latest version.
This would break consumers of this library.

SONAME should be something like ``````libuhdm.so.A.B.C```, with numbers A, B, C.

timkpaine:
This PR does two things:
- instead of installing into `lib/uhdm/`, we will install into `lib`. This is the normal method, installing into `lib/uhdm/` has caused problems on both conda and homebrew, and is sure to continue to cause headaches
- we use `SOVERSION` so that `lib/libuhdm.so` is now a symlink, and we install a versioned `lib/libuhdm.so.<major>.<minor>`

Fixes #992 

(I will do a similar PR for Surelog)

---

```
-- Installing: /tmp/uhdm-install/lib/libuhdm.so.1.71
-- Installing: /tmp/uhdm-install/lib/libuhdm.so
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ endif()
 
 # NOTE: Set the global output directories after the subprojects have had their go at it
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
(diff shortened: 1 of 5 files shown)


alaindargelas:
@hzeller please review early next week.

mithro:
@hzeller is out of office until 28th August

timkpaine:
there's no explicit rush (other than the desire to not leave PRs open for too long).

PR #1005, opened 2023-08-28
Make sure serializers include the config.h

hzeller:
The files should be self-contained and include the header that defines the windows-specific O_BINARY.
(Currently that only works accidentally if compiled with cmake)

### DIFF
--- a/templates/Serializer_restore.cpp
+++ b/templates/Serializer_restore.cpp
@@ -44,6 +44,7 @@
 #include "UHDM.capnp.h"
 #include <uhdm/uhdm.h>
 
+#include "uhdm/config.h"
 
 namespace UHDM {
 static constexpr std::string_view kUnknownRawSymbol = "<unknown>";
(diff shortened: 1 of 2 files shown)


hzeller:
Compilation units should always be self-contained and include everything it needs, in particular when it is to be distributed - you can't control how people will compile a library, so force-includes on the command line is not really used for that.

hzeller:
> But, that's exactly how config.h is included in the project. 
> None of the files include config.h explicitly. This is not limited to serializer only implementation. 

Right, but only the serializers are using it currently, that is why I included it there. It really should be included in all places.

> Same is true for Surelog project as well.

Yes, I noticed that, and I will fix that in a later PR there :)
 
> I agree that headers should be explicitly included 

Yes. Let's apply it always :)

> but, in my opinion, configuration headers are out of that scope. 

Including config headers is the only way to guarantee what we need. So it should also be considered in the same scope.
For the reason that you can't enforce any way for users of the library to use that particular way, this is needed.

> On top that, config.h doesn't actually exist. It is generated by CMake, and so any build system will have to do something similar.

config.h is generated and will be distributed along with the other headers, so it will work well when included in files.
Requiring to include via the command line is somewhat orthogonal to that.

PR #1018, opened 2023-09-13
Use ref_obj to reference typespecs

hs-apotell:
Use ref_obj to reference typespecs

scope::Typespecs are cloned from non-elaborated tree to the elaborated tree. However, at the moment, this collection doesn't have all the typespecs in the non-elaborated tree.

This change solves a few different problems -

* Avoids too many duplicates in the elaborated tree.
* ref_obj will hold the type location information in parameter declaration.

Known Issues -
* Need to collect all typespecs in the scope subtree to populate scope::Typespecs
* Though the typespecs are cloned, elaborated tree is still using the ones from the non-elaborated tree. The tree is still complete but typespecs are crossing the boundary between elaborated and non-elaborated. Need to fix this during binding.

### DIFF
--- a/model/array_typespec.yaml
+++ b/model/array_typespec.yaml
@@ -39,12 +39,12 @@
   - class_ref: index_typespec
     name: index typespec
     vpi: vpiIndexTypespec
-    type: typespec
+    type: ref_obj
     card: 1
   - class_ref: elem_typespec
     name: elem typespec
     vpi: vpiElemTypespec
-    type: typespec
+    type: ref_obj
     card: 1
   - obj_ref: resolution_func
     name: resolution func
--- a/model/bit_typespec.yaml
+++ b/model/bit_typespec.yaml
@@ -24,7 +24,7 @@
   - obj_ref: bit_typespec
     name: bit typespec
     vpi: vpiElemTypespec
-    type: bit_typespec
+    type: ref_obj
     card: 1
   - obj_ref: ranges
     name: ranges
@@ -44,7 +44,7 @@
   - class_ref: typespec
     name: typespec
     vpi: vpiIndexTypespec
-    type: typespec
+    type: ref_obj
     card: 1
   - property: sign
     name: sign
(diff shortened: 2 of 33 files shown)


model/array_typespec.yaml
```diff
@@ -39,12 +39,12 @@
   - class_ref: index_typespec
     name: index typespec
     vpi: vpiIndexTypespec
-    type: typespec
+    type: ref_obj
```
alaindargelas, line 42: That's not good.
Within a typespec there should be a strong type, at least it should be a ref_typespec.
Actually everywhere it should be a ref_typespec, you should not have used ref_obj.

model/bit_typespec.yaml
```diff
@@ -24,7 +24,7 @@
   - obj_ref: bit_typespec
     name: bit typespec
     vpi: vpiElemTypespec
-    type: bit_typespec
+    type: ref_obj
```
alaindargelas, line 27: Here it is even worse from a specific bit_typespec we now have a ref_obj to anything.
There should be a ref_typespec whose only choices are typespec classes

alaindargelas (changes requested):

model/bit_typespec.yaml
alaindargelas: @hs-apotell , I'm sorry, I'm not sure what was I thinking back then. 
We use ref_obj in too many places now, it is almost like a void* (any).
I think now we should go with https://github.com/chipsalliance/UHDM/pull/984

alaindargelas: Done. Please keep for other PRs all the changes that are not of the matter at hand.

alaindargelas: Isn't the point of ref_typespec to be a pointer to an actual typespec object that can be shared by multiple ref_typespec?
Else you clone all typespecs which will have a huge memory impact.

alaindargelas: We should have a field actual_typespec of type typespec.
We need to have an explicit pointer from ref_typespec to a a typespec

alaindargelas:
@hs-apotell , we can close this PR now?

PR #1019, opened 2023-09-15
do not return early from reportError

wsipak:
After a recent change to `synth subset checker` we're not able to check if a node belongs to the non-synthesizable set or not, and as a result non-synthesizable parts of this test cause errors:
https://github.com/zachjs/sv2v/blob/764a11af7f861dfbf9e2c40422fe7b02f7a166c2/test/core/struct_part_select.sv

Here's the change in UHDM:
https://github.com/chipsalliance/UHDM/commit/da7e6cfc0d84a3498f8a90ec050f5d57cd128b41

I've observed that we can safely bump the Surelog submodule when we use UHDM without the early return from `SynthSubset::reportError`
I've run tests here: https://github.com/chipsalliance/synlig/actions/runs/6198609315/attempts/1

FYI @kbieganski

### DIFF
--- a/templates/SynthSubset.cpp
+++ b/templates/SynthSubset.cpp
@@ -98,7 +98,6 @@ void SynthSubset::report(std::ostream& out) {
 }
 
 void SynthSubset::reportError(const any* object) {
-  if (isInUhdmAllIterator()) return;
   const any* tmp = object;
   while (tmp && tmp->VpiFile().empty()) {
     tmp = tmp->VpiParent();


alaindargelas:
The early return was put in place for non-elab models which sometime contain non-synthesizable constructs.
But only the elaborated model should be our concern.
Lots of code is ignored in the elaborated model and in a lot of designs the non-synthesizable code is filtered out.
That is why it is important that this check only runs on the elaborated code.
This change you are proposing is going to break Surelog regressions and tests (Running outside of these repo) that will not synthesize any more.

@wsipak, @kbieganski I think the real issue is the plugin is still looking at the non-elaborated tree where it should not.
It should only navigate the vpiTopPackage and vpiTopModule iterators and not the vpiAllModule, nor vpiAllPackage

alaindargelas:
Actually, I forgot that I implemented another filter, if a module is not part of the design with -synth, I remove the module from the allModules already, so we can accept this PR.

wsipak:
@alaindargelas thank you very much. This will make it easier to sync the plugin with the recent version of Surelog. However, we also need a bump of the UHDM submodule in Surelog itself.

alaindargelas:
@wsipak. Done: https://github.com/chipsalliance/Surelog/pull/3864

PR #1020, opened 2023-09-16
Sort capnp factories, ref_typespec, and naming consistency

hs-apotell:
Sort capnp factories, ref_typespec, and naming consistency

* Sort the factories in capnp output so comparisons are easier
* Introducing ref_typespec, parallel to ref_obj to reference typespecs
* Enforce naming consistency


alaindargelas:
@hs-apotell can this be merged before https://github.com/chipsalliance/UHDM/pull/1018 which is a work in progress?

PR #1081, opened 2024-07-30
Fix compilation with gcc 14; Fix evaluation of `$left`, `$right`, `$high`, `$low`

kamilrakoczy:
This PR fixes compilation for newer gcc version by adding missing include.

It also fixes evaluation of `$left`, `$right`, `$high`, `$low` by adding missing `$` when comparing function name.


alaindargelas (approved):

tgorochowik:
Thank you for the review!

PR #1097, opened 2025-03-06
Add missing break in switch.

hzeller:
It accidentally doesn't make any difference, as this is the last case in that switch, but this removes ambiguity about intention.

### DIFF
--- a/templates/NumUtils.cpp
+++ b/templates/NumUtils.cpp
@@ -25,6 +25,7 @@
 
 #include <algorithm>
 #include <bitset>
+#include <cstdlib>
 #include <cstring>
 #include <iostream>
 #include <locale>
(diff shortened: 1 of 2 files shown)


alaindargelas:
@hzeller did you check Surelog regression with this change?
I pulled the last 3 PRs into Surelog and got this:
https://github.com/chipsalliance/Surelog/pull/4026

hzeller:
So the issue on Windows is due to `<chrono>` not included in a antlr4-header, which seems to be fixed upstream in antlr (but that is not released yet). Maybe we can use an intermediate version of antlr from head in our `third_party/` ?

The other failures look weird as they seem to be just not running at all (Mac-os), possibly due to a runner issue, or interrupted after a while (like the regression tests). I've not re-triggered the failed runs to see what happens now.

This particular change and the others I did should be benign (e.g. this on, the break in that switch would've behaved the same as before)

Let's see what the [rerun](https://github.com/chipsalliance/Surelog/actions/runs/13891536737) is doing. I expect windows still to fail due to the antlr issue, but the others should make more progress.

PR #1098, opened 2025-03-07
Fix evaluation of $size for array

Rodrigodd:
Addresses #1094 

The solution is very ad-hoc. A better solution would be to move `$size` implementation from `ExprEval::size` to where `$high` and `$low` are evaluated, but that section of code is also unused, untested and very likely broken. See https://github.com/chipsalliance/UHDM/issues/1094#issuecomment-2705321758 for more details.

### DIFF
--- a/templates/ExprEval.cpp
+++ b/templates/ExprEval.cpp
@@ -1272,7 +1272,9 @@ uint64_t ExprEval::size(const any *ts, bool &invalidValue, const any *inst,
     case UHDM_OBJECT_TYPE::uhdmarray_typespec: {
       array_typespec *lts = (array_typespec *)ts;
       ranges = lts->Ranges();
-      if (const ref_typespec *rt = lts->Elem_typespec()) {
+      if (!full) {
+        bits = 1;
+      } else if (const ref_typespec *rt = lts->Elem_typespec()) {
         bits = size(rt->Actual_typespec(), invalidValue, inst, pexpr, full);
       }
       break;


alaindargelas:
Running workflow

alaindargelas:
@Rodrigodd did you check the Surelog regression with this change?
I pulled the last 3 PRs in UHDM in Surelog, and one of them is the culprit:
https://github.com/chipsalliance/Surelog/pull/4026

Rodrigodd:
@alaindargelas I believe I did, I even add the following tests to Surelog. But maybe I didn't run the right command, or my repository was too dirty. I will run them again on a clean repo to see.

```
$ cat tests/SystemCallSize/dut.sv                                                               0.01s
module top();

logic [3:0] a[1:0];
logic b[$bits(a)-1:0];

logic c[$size(a)-1:0];
logic d[$high(a)-$low(a):0];

endmodule
```
```
$ cat tests/SystemCallSize/SystemCallSize.sl
-parse -d uhdm -d coveruhdm -elabuhdm -d ast dut.sv -nobuiltin
```

alaindargelas:
@Rodrigodd , can you submit a PR in Surelog with your new test (I didn't see a PR) and the UHDM tag update that correspond to your change?
Thanks

Rodrigodd:
@alaindargelas I will submit a PR on monday with the test case. I can rebase it later if you decide it is better to merge to other PRs before that one.

PR #1099, opened 2025-03-19
CI: Update ubuntu to 24.04

hzeller:


hzeller:
... but now that we talked about Ubuntu upgrade on the Surelog side: looks like UHDM CI is in need of an update as well.

PR #1096, opened 2025-02-28
chipsalliance/UHDM#1095: Get implementation of UhdmListener & VpiListener in sync

hs-apotell:
chipsalliance/UHDM#1095: Get implementation of UhdmListener & VpiListener in sync

Issue #1095, opened 2025-02-28
Differences between UhdmListener and VpiListener


mlr11:
I was trying to use UHDMListener and didn't get the behavior I was expecting.  I compared it's behavior to VpiListener and noticed that UHDMListener is visiting some extra nodes that VpiListener isn't.

I traced it to things like this:

from: UhdmListener.cpp
```C++
void UhdmListener::listenExpr_(const expr* const object) {
  listenBaseClass_(object);
  if (const any *const typespec_ = object->Typespec()) {
    listenAny(typespec_);
  }
}
```

from VpiListener.cpp:
```C++
void VpiListener::listenExpr_(vpiHandle handle) {
  if (vpiHandle itr = vpi_handle(vpiTypespec, handle)) {
    listenAny(itr);
    vpi_free_object(itr);
  }
}
```

UHDMListener.cpp does an extra `listenBaseClass_(object);` that isn't present in VpiListener.cpp

Also, the UHDMListener example: uhdm_listener_test.cpp has some issues:

The body of leaveModule_inst will never be executed.
` if (visited.find(object) != visited.end()) return;`

This will always return, the object will always be in the visited container; this is the leave function, the object was added in the enter function.
If you trace this code, you'll see more enter calls than expected (see above) and no leave calls being executed.

PR #1100, opened 2025-03-30
Revert "chipsalliance/UHDM#1095: Get implementation of UhdmListener & VpiListener in sync"

alaindargelas:
Reverts chipsalliance/UHDM#1096 for diagnostic, will remerge if OK.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
 
   macos:
     name: macos-gcc
-    runs-on: macos-13
+    runs-on: macos-12
 
     steps:
     - name: Setup Python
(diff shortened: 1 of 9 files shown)


alaindargelas:
@hs-apotell there are 2 problems mixed up here, the chrono problem on Windows should be addressed when https://github.com/chipsalliance/Surelog/pull/4029 can finally be merged.
But this PR of yours that I reverted breaks every platform tests in Surelog. The failure are not compilation failures, they are functional failures, the test produce different results compared to the logs.

alaindargelas:
@hs-apotell can you resubmit this PR with the corresponding Surelog ref log updates?

PR #1104, opened 2025-05-04
Adding standard UHDM packaging and utility module

dpetrisko:
Hello!

This PR modifies the python packaging flow to a more standard sdist based approach
(https://stackoverflow.com/questions/29232614/build-a-python-package-with-setup-py-in-cmake). This makes it a lot easier to add more functionality to the package. As an example, I also added a generator wrapper to the util package so that you can replace

```
module_iterator = uhdm.vpi_iterate(uhdm.uhdmallModules,data[0])
while(True):
   vpiObj_module = uhdm.vpi_scan(module_iterator)
   if vpiObj_module is None:
       break
   # stuff
```

with 

```
for vpiObj_module in util.vpi_iterate_gen(uhdm.uhdmallModules,data[0]):
   # stuff
```

which cleans up recursive exploration a lot. My plan is to rewrite the uhdm-hier and uhdm-dump utilities in python to demo how to use them, but I'll make a separate PR for this. Let me know if there's anything you need from me. Thanks!


python/tests/test_module.py
```diff
@@ -19,6 +19,17 @@ def test_module(self):
 
         self.assertEqual(set(result),set(["module2","module1"]))
 
+    def test_generator(self):
+        result = []
+
+        s = uhdm.Serializer()
+
+        data = uhdm.buildTestDesign(s)
```
Thomasb81, line 27: If you have an better idea that to use a build-in fake design to non-regress because it appears to be not very representative of data that can be present in uhdm...

Thomasb81: > My plan is to rewrite the uhdm-hier and uhdm-dump utilities in python to demo how to use them

That would be excellent !

timkpaine:
I would consider doing it differently, and honestly its easy enough as its own repo: https://github.com/dau-dev/uhdm-pypi

alaindargelas (approved):

alaindargelas (changes requested):
Actions failed on the python dist step.

dpetrisko:
Took a look at the logs, seems like it was a missing pip dependency. BTW is there a way to run these checks locally?

dpetrisko:
```
D:\a\_temp\msys64\usr\bin\cmake.exe
cmake version 4.0.1
...
CMake Error at third_party/capnproto/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

Very weird, it looks like it's picking up the wrong cmake? Or there's a bug with cmake 4 detection with minimum 3.4?

PR #1154, opened 2026-09-13
ExprEval: nested function calls whose formals share a name recursed until the stack was exhausted

alaindargelas:
## Summary
`outer(inner(x))` with both formals named `g` and **non-constant** actuals crashed Surelog (SIGSEGV, infinite `getValue -> reduceExpr -> getValue`): `evalFunc` reduced the outer argument `inner(x)` to inner's partially reduced body `{g[0], g[1]}` and bound it to outer's `g` **by name**, so `g` in the outer body reduced to an expression containing `g`. OpenTitan `aes_sbox_dom` hits it on `aes_scale_omega2_gf2p2(aes_square_gf2p2(a ^ b))` — every masked DOM S-box (aes_sub_bytes with its default `SecSBoxImpl = SBoxImplDom`).

- `evalFunc`: a non-constant reduced argument that still references one of the function's own formals cannot be bound by name → `invalidValue` (the call is simply not a compile-time value; elaboration keeps the call).
- `reduceExpr`: a thread-local depth guard (1024) turns any remaining self-referential reduction into `invalidValue` instead of a crash.

```systemverilog
package p; function automatic logic [1:0] outer(logic [1:0] g); return {g[0], g[1] ^ g[0]}; endfunction
           function automatic logic [1:0] inner(logic [1:0] g); return {g[0], g[1]}; endfunction endpackage
module m (input logic [1:0] a, b, output logic [1:0] x); import p::*; assign x = outer(inner(a ^ b)); endmodule
```
Renaming inner's formal avoided it; with the fix both spellings elaborate and yosys (read_uhdm) evaluates `x` correctly.

## Test plan
Surelog regression with this UHDM: running now, result posted below. Surelog test `tests/NestedFuncSameFormal` follows in the Surelog bump PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XshwvRJtjR5dA48M4b1CLM


alaindargelas:
Narrowed the evalFunc refusal to arguments that are an operation / select / call referencing one of the callee's formals; a **bare** ref_obj to a same-named formal keeps the historical path (getValue returns the param_assign Rhs without re-reducing it, so it never looped, and the typespec annotations evalFunc leaves on the body stay as before).

Surelog regression with this UHDM: 795 tests, **0 FAIL / 0 SEGFLT**; 9 goldens change — 5 only in the UHDM object-count statistics (fewer temporary objects since refused calls no longer build a scope), 4 (FuncArgs, FuncStruct, PackFuncParent, StringRange) lose a `vpiTypespec` annotation on an argument expression of a call that cannot be constant-evaluated anyway (recursive `pow_a(base, exp - 1)` with unbound `exp`). The refreshed goldens + `tests/NestedFuncSameFormal` come with the Surelog bump PR. The OpenTitan masked DOM S-box (`aes_sub_bytes`) now elaborates, proves equivalent vs read_slang and co-simulates clean.

alaindargelas:
Full Surelog regression with the narrowed fix: **795 tests, 0 FAIL / 0 SEGFLT / 0 DIFF** after refreshing 14 goldens. Two of them turned out to be silent crashes of the same recursion: the `Rggen` and `Earlgrey_Verilator_01_05_21` goldens ended at "Creating UHDM Model..." with no summary (the process had died there); both now run to completion (`[  FATAL] : 0 … [  ERROR] : 0`). The other refreshed goldens change only in object-count statistics or lose a vpiTypespec annotation on a non-evaluable recursive-call argument.

